PR #9, opened 2015-08-05
Added some features

rezaies:
- Improve the template management. It's now possible to change a template and save it with the same name.
- It's now possible to set the an optional margin for individual pages.
- It's now possible to set an optional width limit for each element. This is useful to make sure elements do not collapse.
- Implemented support for element alignment.
- Implemented support for specifying the reference point location for each element. This is the location of the element that's put at posx and posy.
- added drag & drop feature to rearrange certificate elements


markn86:
Hi Shamim,

Thanks for all your work on this. I do like the look of it all.

I just have a question regarding the addition of the margin attribute - why exactly is this required and if it was to be introduced, why would we not add a left margin as well?

Thanks.

rezaies:
Hi Mark,

Suppose that you are designing a certificate page with many elements. you usually know where your element begins from, but you don't know where they end. You don't want the element texts to get too close to the right edge of the page. It's possible to set a max width for all your elements, but it's hard and also needs some mathematical, although basic, calculations. It would be much easier to just set a right margin for the whole page and not be bothered to set max width for almost all elements.

The whole thing I said above was valid until I added the "reference point location" feature. After introducing that, I think it's also good to have the left margin. because if an elements's reference point location is "top right"....., we know where the right edge of the element is, but we don't know where the left edge might be. so it's good to have left margin as well.

markn86:
Thanks for the explanation!

Sorry for taking so long getting to this. I broke my collarbone and required surgery, so haven't been using my computer much.

I have taken your commits and did some changes to the wording, and added some of my own. It's now on master.

Thanks!

PR #73, opened 2017-01-24
Update customcert.php

scyrma:
Add missing string in customcert.


markn86:
Cheers, will get around to cherry-picking this shortly.

markn86:
Sorry, this does not fix the issue. The string should be 'customcert:viewallcertificates', but thanks for pointing it out.

Issue #89, opened 2017-03-03
Some images do not appear when generating PDF


scyrma:
The background picture element works fine in the "reposition" view, but the image does not appear in the final PDF.

markn86:
Yep, can confirm this is an issue. Working on fix now, cheers!

markn86:
I can only replicate this issue if I use a .svg. Was this the filetype you were using as well or was it another? Thanks.

markn86:
The reason I ask is I would like to create a fix for all file types, not just one. What would be even better is if you attached the image you were trying to use! :)

scyrma:
Yes, i think it was an svg.

crmpicco:
I appear to be having the same problem with the PNG below.

![border](https://cloud.githubusercontent.com/assets/2536923/23645449/1545390c-0347-11e7-82d5-c63e48ec441e.png)

markn86:
Thanks for that.

This is also an issue with the 'image' element. However, you can ALSO get the TCPDF error 'Unable to get the size of the image' if you do not specify a width or height for the image. The 'bgimage' element uses the width and height of the PDF so you don't get this error, but the 'image' element attempts to calculate the width and height if it isn't specified resulting in the following -

![photo_2017-03-07_15-52-22](https://cloud.githubusercontent.com/assets/783213/23646909/1688528e-034e-11e7-9828-2cd13ff2c4a6.jpg)

crmpicco:
OK, cheers for the explanation. :+1:  I can see that being a problem for cloud users as they may not always enter the image dimensions, or know where to find them.

markn86:
Yep. And even if they do enter it the image won't be displayed. So, it's something that needs fixed.

markn86:
The image supplied works fine for me.

crmpicco:
Does it display on the PDF?

markn86:
Yep. And so does the watermark.png image from the MDL tracker issue.

crmpicco:
We are also use TCPDF 6.2.12.
```
Name: TCPDF
Version: 6.2.12
Release date: 2015-09-12
Author:	Nicola Asuni
```

markn86:
Discovered this was due to a custom file system being used on Moodle cloud. 

SVG images will still be a problem but will deal with that in a separate issue.

markn86:
Thanks all, closing!

PR #91, opened 2017-03-08
#89 Copy files to temp directory before accessing

andrewnicols:
Note: I haven't tested this yet.


markn86:
Thanks a lot Andrew, appreciated!

PR #95, opened 2017-03-15
Changed the way style is generated

ovidiumorariu:
When the style is generated if you use ":" it will write in a single line font-family, font-weight and font-style, using ";" it will write each style element in a different line.


markn86:
Thanks! Cherry-picked into master, MOODLE_32_STABLE and MOODLE_31_STABLE.

Issue #100, opened 2017-04-24
Add SVG support for image and bgimage elements.


Dave-B:
Currently, only the image and bgimage elements only use TCPDF's [Image()](https://tcpdf.org/docs/source_docs/classtcpdf/#a714c2bee7d6b39d4d6d304540c761352), which only supports raster formats.

TCPDF also provides [ImageSVG()](https://tcpdf.org/docs/source_docs/classtcpdf/#a56536508fb1b5aede7d2ed27f56c2353) for SVG support.

Pull request coming soon...

Support for EPS/AI _could_ be added via [ImageEps()](https://tcpdf.org/docs/source_docs/classtcpdf/#ace0583441eed657b827420fcfb37a3bd), but I don't think it's worth adding - I've not had great experience with it, due to its limited EPS compatibility.

markn86:
Thanks Dave!

I recently noticed this support in TCPDF when fixing another issue but it escaped my attention and I never ended up creating a follow-up issue. So, thanks for doing that for me!

When creating your pull request please prepend the issue number (#100) before the commit message.

Thanks again!

markn86:
(will add the issue number just letting you know for future sake :))

markn86:
Done! 👍 Thanks again.

PR #101, opened 2017-04-24
Add SVG support for image and bgimage elements.

Dave-B:
Fix for  #100.

Tested on mod_customcert MOODLE_31_STABLE.
Applies cleanly to master and MOODLE_32_STABLE, and I expect it should work ok on all those branches.


Dave-B:
Actually, I've just noticed that this breaks the HTML preview for "Reposition elements". I'll investigate...

markn86:
Thanks Dave, merging. :)

Dave-B:
Cool. Thanks for the plugin BTW :-)

PR #147, opened 2017-11-20
Resolves PHP coding style problem.

michael-milette:
Hi @markn86 ,

I noticed you have some code prechecks issues on Moodle.org, especially one issue because verify_certificate.php file doesn't have anything to authenticate users or prevent them from accessing the page.

I realize this is by design. I had the same problem with one of my plugins. I found that one way to get around it was to add three lines near the top of the source code:

if (false) { // This is only included to avoid code checker warning.
    require_login(); // We don't ever actually want to require users to be logged-in.
}

So that is what this pull request is all about. Just to eliminate the code precheck error. As you can see, it does absolutely nothing but is enough to make the code prechecker happy.

Thanks for making this plugin available to the Moodle community.

Best regards,

Michael Milette

### DIFF
--- a/verify_certificate.php
+++ b/verify_certificate.php
@@ -24,6 +24,10 @@
 
 require_once('../../config.php');
 
+if (false) { // This is only included to avoid code checker warning.
+    require_login(); // We don't ever actually want to require users to be logged-in.
+}
+
 $contextid = optional_param('contextid', context_system::instance()->id, PARAM_INT);
 $code = optional_param('code', '', PARAM_ALPHANUM); // The code for the certificate we are verifying.
 


markn86:
Hi Michael,

This has been bugging me for some time now.

I am going to ask a few other developers at HQ to see if there is any other way to get the codechecker to stop complaining, if not will cherry-pick this commit and backport to stable branches.

Get back to you soon.

markn86:
For reference https://moodle.org/mod/forum/discuss.php?d=355015#p1432027 and https://github.com/moodlehq/moodle-local_moodlecheck/issues/38.

michael-milette:
Thanks @markn86, that is an excellent reference. Much appreciated and will be useful.

markn86:
No problem Michael. Closing this issue now.

markn86:
(The last link is to show that it doesn't work in all cases).

PR #176, opened 2018-04-10
only getting users details who are enrolled within the course

tristanmurdoch:
Hi markn86
We are currently deploying custom cert and during testing it was discovered that when emailing the certificate and with Email teachers = true, we would also email to admin support users (not System Admin) who had no enrolment within the course.

We have several system level roles with course:update permission allowed for users ( Admin Support Staff ). 
As you are getting users with that capability on a site level, they are included in the email list.
If we limit this to just enrolled users, it will only send to the required people for that course.

Any reason you used get_users_by_capability?

Cheers
Tristan

### DIFF
--- a/classes/task/email_certificate_task.php
+++ b/classes/task/email_certificate_task.php
@@ -70,8 +70,7 @@ class email_certificate_task extends \core\task\scheduled_task {
 
                 // Get the person we are going to send this email on behalf of.
                 // Look through the teachers.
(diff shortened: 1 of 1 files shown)


markn86:
Apologies, this escaped my attention. I have been buried in the 'Issues' tab on github and managed to forget about 'Pull requests'. I have made a note to look at this when I start more development on this project.

markn86:
Sorry for the delay, this has been cherry-picked! Thanks!

PR #187, opened 2018-04-30
Fix ORA-00932 inconsistent datatypes

mwithheld:
Fix ORA-00932 inconsistent datatypes error.  3 is the minimum email address length (a@b).


markn86:
Thanks for creating this.

We should actually be using `$DB->sql_length` to ensure it is cross-db compatible. For example, in mysql we want CHAR_LENGTH.

markn86:
I created a fix. Closing, thanks!

PR #192, opened 2018-05-24
Fixed report fields for download

bobopinna:
This patch fixes report download, removing user pix and actions column from reports.


bobopinna:
This commit apply to brances from MOODLE_31_STABLE to MOODLE_35_STABLE

markn86:
Whoops, should also say thanks for the report and patch! :)

markn86:
I went ahead and backported it, but it won't be available in the plugins DB until next release. Cheers!

Issue #197, opened 2018-06-10
My Certificates and Verify Certificate pages not processed through Moodle filters


michael-milette:
Course name and Certificate name are not processed through Moodle filters on the **My Certificates** page, both in the listing and when you download the table data page, and the **Verify Certificate** page when a code is found to be valid. As a result, multi-language course names and certificates are not displayed correctly.

Note: Pull request coming shortly to correct these issues.

michael-milette:
See Pull Request #198 .

markn86:
Created a fix. Will be available next release. :)

michael-milette:
Thanks you Mark! 👍

PR #198, opened 2018-06-10
Added format_string() to My Certificates/downloads and Verify certification pages

michael-milette:
This fix addresses the issues described in project issue #197.

Please note that the following 2 lines from customcert/my_certificates.php also had to be moved up a few lines because Moodle was complaining that the $PAGE->context was not set when downloading the data from the My Certificates page:

    $PAGE->set_url($pageurl);
    $PAGE->set_context(context_user::instance($userid));

Let me know if you have any questions or concerns.

Best regards,

Michael Milette

### DIFF
--- a/classes/my_certificates_table.php
+++ b/classes/my_certificates_table.php
@@ -94,7 +94,7 @@ class my_certificates_table extends \table_sql {
      * @return string
      */
     public function col_name($certificate) {
-        return $certificate->name;
+        return format_string($certificate->name);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


markn86:
Thanks Michael.

It is good practice to pass the context to the format_string call instead of relying on $PAGE->context (which it does by default). This can have some nasty consequences in the future, albeit in this case since this is unlikely ever to be called by CLI it's rare, it's still something I do. I am going to create a new patch that passes the context to these functions, but thanks for bringing this to my attention.

markn86:
I thought about this more and you probably don't want to rely on ```$PAGE->context``` as it could be the user/system context and we would want to use the course module context to ensure the correct filters are used.

michael-milette:
Thanks for the feedback Mark! I really do appreciate it. 👍

Issue #213, opened 2018-07-17
Allow element plugins to have admin settings 


dmitriim:
At the moment there is no admin tree for subplugins. Need something similar to assign plugin.

dmitriim:
Pull request for this issue https://github.com/markn86/moodle-mod_customcert/pull/214

markn86:
Thanks for your contribution, much appreciated. I have cherry-picked this commit. Closing!

PR #214, opened 2018-07-17
Allow element plugins to have admin settings for fixing #213

dmitriim:


markn86:
Done as per comment in #213. 👍

PR #216, opened 2018-07-18
Add date range element to close #185

### LINKED ISSUE #185: Add a daterange element subplugin
In some scenarios, you might need to create a certificate about a given lecture term, when it's not important at which day it was issues but in which term it was issued.

We propose to add a new datarange element subplugin, based on the date element subplugin, which automatically shows the configured daterange string from the configures daterange.

For example:
* The teacher adds the daterange element to his template
* He configures the daterange string to be displayed:
** From 1st april to 30th september: String is "summer term"
** From 1st october to 30th march: String is "winter term"
* The teacher can still use the "date item" setting where he can choose if the issued date, the course start date and so on should be used.
* Assumed that date item is set to "issued date": If the student downloads the certificate at 5th may 2018, the daterange element will output "summer term 2018".

dmitriim:


dmitriim:
This new element could use benefits of having admin settings from https://github.com/markn86/moodle-mod_customcert/pull/214

element/daterange/styles.css
```diff
@@ -0,0 +1,3 @@
+.datestring {
+    padding-bottom: 1.5rem!important;
```
abias: 1. In my test setup, this css rule wasn't delivered to the browser at all. I assume that styles.css files aren't processed in subplugins of this plugin.
2. After added to the page manually, there was no difference when the rule had !important set and when it had not. I assume that !important would be overkill here.
3. I understand why you added this rule, but I am afraid that it does not improve the form. I would rather propose to omit this rule for now.

dmitriim:
Hi @markn86 !
I think @abias is happy with everything in this pull request. Could you please review this feature and merge it in. Thanks!

markn86:
Thanks, will look at this when I get some free time. Appreciate the contributions!

abias:
Hi Mark, 
please hold your review one more time. I have found another gap in my specification which I have to discuss with Dmitrii.

element/daterange/classes/element.php
```diff
@@ -262,15 +268,22 @@ public function validate_form_elements($data, $files) {
...
             $enddate = $this->build_element_name('enddate', $i);
+            $rangeperiod = $data[$enddate] - $data[$startdate];
 
+            // Check that end date is correctly set.
             if (!empty($data[$enabled]) && $data[$startdate] >= $data[$enddate] ) {
                 $errors[$this->build_element_name('enddate', $i)] = get_string('error:enddate', 'customcertelement_daterange');
             }
+
+            // Check that recurring dateranges are not longer than 12 months.
+            if (!empty($data[$recurring]) && $rangeperiod > self::MAX_RECURRING_PERIOD ) {
```
abias: I think this should be >= here instead of >.
Otherwise, I would be able to define a daterange from for example 1st october 2017 to 1st october 2018 which is in fact more than the 12 months limit

element/daterange/lang/en/customcertelement_daterange.php
```diff
@@ -33,7 +33,7 @@
...
+$string['help'] = 'Configure a string representation for each daterange. Set start and end dates as well as a string you would like to transform each range to. Make sure your ranges do not overlap, otherwise the first detected daterange will be applied. If no daterange applied to a date, then Fallback string will be displayed. If Fallback string is not set, then there will be no output at all. If you mark a date range as Recurring, then the configured year will not be considerred and the current year will be used. Also there are {{first_year}} and {{last_year}} and {{current_year}} placeholders that could  be used in the string representation. The placeholders will be replaced by first year or last year in the matched range or the current year.';
```
abias: There's a typo: "If no daterange _applies_ to a date, then Fallback string will be displayed"

And there's one more typo: "If you mark a date range as Recurring, then the configured year will not be _considered_ and the current year will be used."
After this sentence you should add: "As the year of a recurring date range is not considered, you are not allowed to configure a recurring date range with more than 12 months as it would become  ambiguous otherwise".

dmitriim:
@abias  I have fixed all your feedback comments. @markn86 please review.

abias:
@markn86, as you have seen, @dmitriim has made some more changes to the code based on my internal feedback to him. But I am happy to tell you that we are now done and would really appreciate your integration review.

abias:
@markn86 , now that Moodle 3.6 Core is out of the door, do you have any schedule for us until you will be able to review and integrate this patch?

markn86:
I will be adding this either today or next Monday and then will release new version for 3.6, 3.5, 3.4 and 3.3.

element/daterange/settings.php
```diff
@@ -0,0 +1,40 @@
...
+defined('MOODLE_INTERNAL') || die('Direct access to this script is forbidden.');
+
+$options = [];
+
+foreach (range(1, 365) as $number) {
+    $options[$number] = $number;
+}
+
+$settings->add(new admin_setting_configselect(
+    'customcertelement_daterange/maxranges',
```
markn86: Why do we have this setting? Couldn't we just let the user add/remove them dynamically without a site setting being required?

markn86: Or let the admin just add multiple elements?

markn86:
Hey guys,

Thanks very much for working on this.

I have to say though I am extremely confused and don't really know how it's supposed to work. I don't see the relation between the 'Date item' and the ranges. The ranges only seem to include the year, so is there a reason it is a date_time_selector? I really don't think there should be an admin setting to determine how many date ranges are shown for every single person. Could you not just add multiple 'Date range' elements?

I would like to set-up a video/audio meeting with @abias so I can get an understanding of what is trying to be achieved here because currently I am left in a very confused state. Sorry!

element/daterange/settings.php
abias: @markn86, as far as we know, there is no functionality and no real best practice in mform which would have supported us in building a feature to add dateranges dynamically. So we added an admin setting to define how many dateranges are available in a daterange element. 

In our local setup, this setting is set to 2 as we only want to support dateranges for winter term and summer term. Others may want to set it to 10 or something like that to support up to 10 dateranges.

If you have a good proposal how to add dateranges dynamically, I think I would not mind if you change this particular code.

abias:
Hi @markn86 ,

> I would like to set-up a video/audio meeting with @abias so I can get an understanding of what is trying to be achieved here because currently I am left in a very confused state. Sorry!

I am very sorry that this pull request confuses you. This shouldn't have happened. I will ping you in Telegram to arrange a call to sort out your questions.

Thanks,
Alex

element/daterange/settings.php
markn86: I think the setting needs more definition. I mean 365 date ranges? I doubt that would ever happen, and if it did it would be a messy looking certificate haha. Could you not just keep the element to one date range and if you needed two you could add two of the elements? Perhaps I am missing some functionality where adding two date ranges in one element does something that adding two elements would not.

abias:
Hi @markn86 ,

as we had a call right after the christmas break and as you told me that you would be finishing this off soon, I just wanted to ask you about the status and if there is anything we can to do to have this merged soon.

Thanks,
Alex

markn86:
I hope to get a look at it this week.

element/daterange/settings.php
markn86: If you create a choice activity in a course, there is a button to add more options. I suspect this could be done here. Basically you would have a button to add a new date range, which would save the current element and then redirect to a page where they would be editing the newly created date range. It's not ideal as it creates the date range element before it has been finalised but I think relying on an admin setting is counter-intuitive. This would also mean having a button to remove a date range. Something similar is already done in the custom certificate when you 'add' a new page. What are your thoughts?

abias: Well, your hint with the choice activity was good. I saw that the button which you have referenced is built in https://github.com/moodle/moodle/blob/master/mod/choice/mod_form.php#L70-L71 and there seems to be a possibility to repeat elements in mforms which I simply wasn't aware of.

In choice activity, there isn't a button to remove choices and I think it would be tricky then to built it for daterange elements. However, in the choice activity, the form seems to be smart enough to only store these repeated form element groups which have been filled. If I create 5*3 form elements by clicking the button multiple times, but fill only four elements, 2*3 elements are presented when I open the form again for editing.

Summing up, we can try to have a look at changing the current admin setting to the "add element" button which you proposed, but there won't be a remove element button. Would this be also ok for you?

markn86: Yep, that sounds fine Alex!

dmitriim: @markn86 @abias  I have had a quick look at the code and it seems like repeat_elements method belongs to moodleform class, but we pass just $mform object to render_form_elements method so we can't in the current implementation handle repeated elements. 

@markn86 any suggestions about improving the plugin API to be able to create repeating elements? e.g add a second parameter to render_form_elements as pass edit_element_form instance?

markn86: The $mform object is an instance of `edit_element_form` which extends `moodleform` so it should be available. I should use type hinting to make it obvious.

dmitriim: @markn86 are you sure? It seems like $mform is MoodleQuickForm.

markn86: `class edit_element_form extends \moodleform {` - is `moodleform` not what you are after?

dmitriim: It does, but we pass just $mform to the element method. See https://github.com/markn86/moodle-mod_customcert/blob/MOODLE_36_STABLE/classes/edit_element_form.php#L67

markn86: Ahh. My mistake. Grr. I should be passing `$this`. I would like to change this but am afraid of breaking third-party elements. How about copying the code and putting it in the element helper class? Not ideal, and it's a hack but think it's better than changing what we pass to `render_form_elements` because even if we make the second parameter optional PHP will complain (see https://3v4l.org/N0p4K).

dmitriim: What about modifying get_element_instance method and set edit_element_form there?  We will introduce a new variable in the abstract class and will have set_element_form(edit_element_form $elementform) method to set it. Then in the factory we will set it after we initialise the element class.

markn86: The approach sounds valid. The code should be pretty quick to write so i'll have a better idea if I could review it, but it sounds better than my suggestion of copy + pasting code that we will need to maintain.

Side note - looking at the code there are many things I want to re-write/refactor. The joys of software development.

dmitriim: Just to clarify, is this the only thing that stops this code from being integrated?

markn86: Yes.

I will integrate and then get a UX person to take a look and add issues later to improve it if there are any to do, but would be good to get the bulk of it in.

dmitriim: Hi @markn86 Could you please have a look at the code I have just pushed? It's in this commit https://github.com/markn86/moodle-mod_customcert/pull/216/commits/87bda096a4b3ee2e05a9c70864dd58fbef28d0c5

Issue #180, opened 2018-04-19
Add possibility to set default values of activity instance settings


abias:
When adding a mod_customcert instance to a course, a teacher is offered several options:
* Email students
* Email teachers
* Email others
* Allow anyone to verify a certificate
* Required minutes in course
* Set protection 

The defaults of these options are hard coded.

We propose to add an admin setting for each of the options listed above to let admins set the default values for the activity instance settings. These admin settings would ideally also contain the "Optional" switch to let admins hide certain activity instance settings behind a "Show more" link. The default for these admins settings should reflect the current hard coded values to retain the current functionality of the plugin.

dmitriim:
Pull request for this feature https://github.com/markn86/moodle-mod_customcert/pull/217

PR #217, opened 2018-07-19
Add possibility to selectively disable activity instance settings

dmitriim:


dmitriim:
To this pull request I have also added fix for #180 as it's easier to apply these two features together

mod_form.php
```diff
@@ -55,33 +55,80 @@ public function definition() {
...
+            $mform->setDefault('protection_copy', get_config('customcert', 'protection_copy'));
+            $firstoption = empty($firstoption) ? 'protection_print' : $firstoption;
+        } else {
+            $mform->addElement('hidden', 'protection_print', get_config('customcert', 'protection_print'));
+            $mform->addElement('hidden', 'protection_modify', get_config('customcert', 'protection_modify'));
+            $mform->addElement('hidden', 'protection_copy', get_config('customcert', 'protection_copy'));
+        }
+        $mform->setType('protection_print', PARAM_INT);
+        $mform->setType('protection_modify', PARAM_INT);
+        $mform->setType('protection_copy', PARAM_INT);
```
abias: It probably won't make a huge difference, but would it make sense to change these three lines from PARAM_INT to PARAM_BOOL?

mod_form.php
```diff
@@ -55,33 +55,80 @@ public function definition() {
...
-        $mform->addElement('selectyesno', 'emailstudents', get_string('emailstudents', 'customcert'));
-        $mform->setType('emailstudents', 0);
-        $mform->addHelpButton('emailstudents', 'emailstudents', 'customcert');
+        if (has_capability('mod/customcert:manageemailstudents', $this->get_context())) {
+            $mform->addElement('selectyesno', 'emailstudents', get_string('emailstudents', 'customcert'));
+            $mform->setDefault('emailstudents', get_config('customcert', 'emailstudents'));
+            $mform->addHelpButton('emailstudents', 'emailstudents', 'customcert');
+            $firstoption = 'emailstudents';
+        } else {
+            $mform->addElement('hidden', 'emailstudents', get_config('customcert', 'emailstudents'));
```
abias: I am wondering if adding the setting as hidden element with the default value is correct in all cases.

Imagine this chain of actions:
* User A has mod/customcert:manageemailstudents, user B has not
* User A adds a new instance of mod_customcert to the course and sets this setting to a value which differs from the default.
* User B edits the existing instance and saves the form.
* Then, if I am not wrong, User B will have changed the setting back to the default although he does not have the capability to change the setting.

Do we need to differentiate if the instance will be created or has already been created before and, in the second case, set the hidden element to the existing instance setting instead of to the default?

dmitriim: Very good point! Will fix it.

db/access.php
```diff
@@ -96,4 +96,70 @@
             'manager' => CAP_ALLOW
         )
     ),
+
+    'mod/customcert:manageemailstudents' => array(
+        'riskbitmask' => RISK_XSS,
```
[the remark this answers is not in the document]
dmitriim: yeah, Copy paste issue here...

mod_form.php
dmitriim: Yes, this sounds sense. will change it

mod_form.php
abias: Thanks for confirming my thoughts :)
To make it really bullet-proof, you could also check the capabilities again within data_postprocessing()

dmitriim: Actually I have just tested it and looks like all works good. Default values get saved just for User B when he created a new activity.

abias:
@markn86:
As you see, there has been some refinement and discussion of the submitted code. @dmitriim and me are now done and it's up to you to review the pull request.

Thanks for your work!
Alex

markn86:
This looks really good guys, thanks! 

A couple of issues -

1. It's possible to edit the DOM and change the value of the hidden fields and bypass the capability check. I would suggest we simply remove the 'else' in those capability checks and not bother having them hidden at all, then add the logic to ensure it isn't be altered maliciously via the function `data_postprocessing`.
2. With the current logic you may accidentally overwrite an existing value. If a user with the capability to change 'emailteachers' (for example) and creates a certificate with this set to 'Yes', then someone who does not have this capability edits the certificate (and the default is 'No') then this would overwrite it.

dmitriim:
@markn86 thanks for your feedback. 

1. I'm not sure if this is not overkill as potentially all hidden values could be changes in the same way. Anyway I'll implement this code. Shouldn't take too long.
2. Looks like this is not a case and required data gets saved correctly.

dmitriim:
@abias  @markn86  I have pushed code updates to fix review notes.
Sorry I messed with the code and had to rebase everything and push with "force" option. So probably you would need to check out the branch again.

dmitriim:
Actually please hold on review. There is a bug in the code.

markn86:
Hi Dimitri,

> I'm not sure if this is not overkill as potentially all hidden values could be changes in the same way. Anyway I'll implement this code. Shouldn't take too long.

Then there isn't really a point having the capabilities if it can be bypassed. :)

> Looks like this is not a case and required data gets saved correctly.

Are you sure? If the default value is '0', but someone set it to '1' (who had the capability to change it) and then someone else (who doesn't have the capability) was to edit it and it was a hidden field with value '0' it would surely overwrite the '1' - same situation as editing the DOM element above.

dmitriim:
@markn86 I think now code is ready. Please have a proper look.

markn86:
Cherry-picked and pushed, thanks!

PR #226, opened 2018-08-08
Allow element plugins to control if they can be added to a certificate

### LINKED ISSUE #225: Allow element plugins to control if they can be added to a certificate
Currently, a user can add any installed element plugin to a certificate.
However, there may be some (custom / bespoke) element plugins installed which shouldn't be added by all users.
Thus, I would propose to allow element plugins to control if they can be added to a certificate

abias:
This resolves #225

### DIFF
--- a/classes/element.php
+++ b/classes/element.php
@@ -353,6 +353,16 @@ abstract class element {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


markn86:
Cherry-picked, thanks!

PR #229, opened 2018-08-23
fixes #218

### LINKED ISSUE #218: TCPDF ERROR: Some data has already been output, can't send PDF file
Hi,
in Moodle 3.3.6, custom certificate 3.3.14, 2017050514, after clicking Save changes and preview there is an error:  exception - TCPDF ERROR: Some data has already been output, can't send PDF file
×Debug info: 
Error code: generalexceptionmessage
×Stack trace:
line 2927 of /lib/tcpdf/tcpdf.php: Exception thrown
line 7643 of /lib/tcpdf/tcpdf.php: call to TCPDF->Error()
line 190 of /lib/pdflib.php: call to TCPDF->Output()
line 311 of /mod/customcert/classes/template.php: call to pdf->Output()
line 233 of /mod/customcert/edit.php: call to mod_customcert\template->generate_pdf()
Any changes in plugin to fix this error?

psywhale:
need to clear buffer before sending a pdf.

### DIFF
--- a/edit.php
+++ b/edit.php
@@ -230,7 +230,7 @@ if ($data = $mform->get_data()) {
 
     // Check if we want to preview this custom certificate.
     if (!empty($data->previewbtn)) {
-	ob_clean();    
+        ob_clean();
         $template->generate_pdf(true);
         exit();
     }
--- a/edit.php
+++ b/edit.php
@@ -230,6 +230,7 @@ if ($data = $mform->get_data()) {
 
     // Check if we want to preview this custom certificate.
     if (!empty($data->previewbtn)) {
+	ob_clean();    
         $template->generate_pdf(true);
         exit();
     }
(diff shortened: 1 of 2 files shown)


psywhale:
For some reason my output buffer always had 3 blank lines.  ob_clean will discard the buffer, so calling it and then generating pdf cleared up this issue for me.

markn86:
Thanks for the patch.

However, I am not convinced this is the correct approach. We want to know if we are incorrectly outputting data before we generate a PDF, this solution will simply just hide that fact. I think we need to find the root cause of this.

psywhale:
I agree, the root cause must be found. In the meantime, I have faculty wanting certs. While the debugger output is nice, it will not display the output buffer with escaped newlines. That may be a patch for the pdflib or tcpdf library.

I may have found lead.. 
`
 Potential coding error: Unexpected output whilst processing AJAX request. This could be caused by trailing whitespace. Output received: '\n', referer: https://moodle.wosc.edu/my/
`

Issue #230, opened 2018-08-24
Allow for separate HTML and plaintext email lang strings


parkeyparker:
I've got a client requirement to include multiple paragraphs in the email that is automatically sent out. It appears that the `emailstudentbody` language string is used for both HTML and plaintext email types. What this means is that using HTML elements such as `<p>` and `<br />` works when the email is displayed as HTML but then the plaintext variant displays the raw HTML code.

It would be useful to either have a method in the user interface to modify the certificate email text on a per certificate basis or a second lang string which will handle plaintext emails (and can make use of `\r\n` or similar for new lines).

I will look into whether I can create a PR to do this but wanted to raise as an issue for reference

Thanks

markn86:
Thanks! A PR would certainly help me. :) I don't see why you couldn't add another string but that's just from a quick assessment of the issue you created.

PR #231, opened 2018-08-24
Added plaintext lang variants for email bodies and updated template

parkeyparker:
This will allow for creating HTML emails whilst still retaining the ability to show a plaintext email without worrying about confusing HTML tags appearing


parkeyparker:
I haven't updated versions or changelogs or anything like that as I wasn't sure of your process for that. This is to fix issue #230

markn86:
Cherry-picked, thanks a lot for your contribution!

Issue #196, opened 2018-06-07
Build 2018051701 displays many elements in the wrong place


JosiahCarberry:
After installing build 2018051701 I note that some, but not all, elements are displayed in different places, generally set much too far to the left. Checking the positioning of the elements, I note that they have not changed. It is just how they are displayed in the pdf output.
I attach before and after images.
![before](https://user-images.githubusercontent.com/13927067/41094254-f310bdb0-6a4d-11e8-9a1a-345b609f5dd3.png)

![after](https://user-images.githubusercontent.com/13927067/41094253-f2a2ebc8-6a4d-11e8-8d5d-1476be5a0cd9.png)

markn86:
Grr. This may be because of #170. Investigating ..

markn86:
Can you tell me what those element types are that moved to the left? Are they all 'text' elements?

markn86:
Are you also able to tell me what version you upgraded from?

markn86:
Ah, I was able to replicate this. Thanks, working on fix now ..

markn86:
Created fix. Thanks a lot for the report! Will be updating version on plugins DB soon.

JosiahCarberry:
It might be moot now, but I upgraded from Build: 2017111301

markn86:
New version is available on the plugins DB now. :) Again, thanks for the report so quickly.

JosiahCarberry:
Fixed now. Thank you!

PR #235, opened 2018-09-25
Resolves #90: Make text element obey line breaks

### LINKED ISSUE #90: The text element does not support multiple lines
When adding a text element with text that appears on multiple lines it ends up displaying the text all on one line, but the user may not want this. We should obey line breaks.

marcusboon:
Not sure why we're using FORMAT_HTML but it looks like FORMAT_MOODLE resolves the issue of it displaying on one line

### DIFF
--- a/element/text/classes/element.php
+++ b/element/text/classes/element.php
@@ -68,7 +68,7 @@ class element extends \mod_customcert\element {
      */
     public function render($pdf, $preview, $user) {
         $courseid = \mod_customcert\element_helper::get_courseid($this->get_id());
-        $text = format_text($this->get_data(), FORMAT_HTML, ['context' => \context_course::instance($courseid)]);
+        $text = format_text($this->get_data(), FORMAT_MOODLE, ['context' => \context_course::instance($courseid)]);
         \mod_customcert\element_helper::render_content($pdf, $this, $text);
     }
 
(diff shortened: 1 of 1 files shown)


markn86:
Thanks a lot for your contribution.

So, it seems the only way (before this patch) was to actually add the tag `<br>` in your text to get it on a new line. Not really intuitive for those who don't know HTML. I am just going to check if this change could possibly break any existing certificates before I go forward.

markn86:
I knew this was familiar. I initially did have it as `FORMAT_MOODLE` but changed it in #196 because it broke existing certificates.

marcusboon:
Interesting that having it as FORMAT_HTML makes it work for other certificate types. I wonder if I can try and reproduce the other kind of certificate to see if it breaks.

markn86:
Sorry, but I am going to have to close this because I can't introduce a change that will break the appearance of the PDF if someone upgrades. That would be disastrous - I know from experience! :D I am starting to think the only easy solution to #90 is to suggest people use `<br />`.

Easy reproduction steps (with the proposed solution).

# Add a text element to your certificate.
# Use the reposition page to put it in the centre.
# Download the PDF.
# Notice it is not in the centre.

marcusboon:
That's cool, I'll take it back and tinker around with it some more, but at least now I know why it's FORMAT_HTML :)

markn86:
I have to admit I had no idea either until I did some investigating.

PR #237, opened 2018-10-05
Reduce code nesting level

mwithheld:
Make the code easier to read and understand


markn86:
Hi @mwithheld,

Thanks for working on this.

I squashed your white space fix and am adding a fix of my own. I am going to remove the boolean return, since the method doesn't specify it returns a boolean (we shouldn't do it just because we are returning early) and also moving a code comment to the appropriate place..

Issue #173, opened 2018-03-27
Possible race condition when issuing certificates


bynare:
The scheduled task retrieves all the issued certificates before starting to process ones that have not been issued yet. Because of this, it is currently possible for a certificate to be issued via the UI (view.php), after the scheduled task has fetched the currently-issued certs, while issuing new certificates.

markn86:
I will be updating the plugin on the plugins DB this week with this fix + GDPR compliance.

PR #239, opened 2018-10-06
Avoid failure if 2 userid-customcertid combos

mwithheld:
We somehow have two results for
SELECT * FROM m_customcert_issues WHERE userid = 20426 AND customcertid = 35

   	ID	USERID	CUSTOMCERTID	CODE	EMAILED	TIMECREATED
1	8127	20426	35	1cxhyyRCB6	0	1535058946
2	8128	20426	35	Tg7yyHT9rh	0	1535058945

which is causing a fatal error preventing sending any certificate emails afterward.  Not sure if this is from a restore or the user double-clicking submit somewhere.

For those who have a similar issue, this fix gets around the problem by choosing the entry with the max datestamp.

### DIFF
--- a/element/date/classes/element.php
+++ b/element/date/classes/element.php
@@ -137,9 +137,12 @@ class element extends \mod_customcert\element {
             // Get the customcert this page belongs to.
             $customcert = $DB->get_record('customcert', array('templateid' => $page->templateid), '*', MUST_EXIST);
             // Now we can get the issue for this user.
-            $issue = $DB->get_record('customcert_issues', array('userid' => $user->id, 'customcertid' => $customcert->id),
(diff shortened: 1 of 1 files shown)


markn86:
Thanks for reporting this.

This may have been due to the problem reported in #173.

I am not a fan of the proposed solution as this should never happen. I think the best thing to do would create an upgrade path that will remove the duplicates. This has been done before. However, it would be great to find the root cause of this so we can be sure this won't just keep happening and then requiring yet another upgrade step and on and on ..

mdjnelson:
AFAICT this has been fixed using an alternative approach. See  #304 and #295.

PR #254, opened 2018-11-20
Matched up prevent x description strings

james-cnz:
The description strings for the settings prevent copy, prevent print, and prevent modify were mixed up when added with:
#180 Add possibility to set default values of activity instance


markn86:
Thanks, cherry-picked! 👍

PR #263, opened 2019-02-01
UI improvements: spacing in web page, move table of issued certificates to separate page

nicoroeser:
At Ulm University, we have improved the UI of this plugin a bit. This pull request contains two small changes:

1. Add a bit of spacing between the button for certificate generation (“View certificate”) and the heading. The page looks nicer then.
2. Move the list of issued certificates to a separate web page. As the list can grow quite long, we like to keep it separate. We have adjusted and added to the Behat test so that it can automatically verify that we have not broken anything.

This code can also be used on top of branch `MOODLE_35_STABLE`.

### DIFF
--- a/classes/report_table.php
+++ b/classes/report_table.php
@@ -169,6 +169,7 @@ class report_table extends \table_sql {
             [
                 'id' => $this->cm->id,
                 'deleteissue' => $user->issueid,
+                'listissues' => true,
                 'sesskey' => sesskey()
             ]
         );
--- a/lang/en/customcert.php
+++ b/lang/en/customcert.php
@@ -109,6 +109,7 @@ $string['landscape'] = 'Landscape';
 $string['leftmargin'] = 'Left margin';
 $string['leftmargin_help'] = 'This is the left margin of the certificate PDF in mm.';
 $string['listofissues'] = 'Recipients';
+$string['listofissueslink'] = 'List of issued certificates';
 $string['load'] = 'Load';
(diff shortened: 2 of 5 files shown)


markn86:
Thanks for creating this.

The report was originally on a separate page and then was intentionally moved back to that one page due to a discussion with @barbararamiro (Moodle UX team). With this patch the pages are so similar except there is no 'View certificate' button. I don't see a huge advantage moving this to another page as it's most likely the only thing the administrator/teacher will want to look at. For example, a teacher isnt going to require to download their certificate often (or at all) so it will always involve two clicks for them to get where they want to go to.

I do like the spacing though and will cherry-pick that commit, but am still not convinced by the other ones.

nicoroeser:
Thanks for your explanation, Mark!

Our original motivation to move the data to a separate page was a report from one of our users: she felt that on opening the certificate activity, the page presented an overwhelming amount of data in the table. And that the activity would feel “less cluttered” if we moved the data out of the way so that the user first needs to decide whether it should be displayed or not.

Another possibility that we see here: there could be a button/link that toggles between “show the list of issued certificates” and “don‘t show it”, and that the setting is saved per user. If you think this can be useful and could go into the plugin, please tell us, as for now we‘ll let this idea rest.

We‘re happy to read that you like to cherry-pick the commit about the spacing, thanks! :-)

nicoroeser:
Mark, shall we close this?

markn86:
Thanks Nico, closing.

PR #275, opened 2019-04-11
Fix #274 Add custom course field element to custom cert.

### LINKED ISSUE #274: add support for 3.7 custom course fields
3.7 is coming with support for custom course fields - it would be really good if we could add support for using custom course fields when generating the certificate!

danmarsden:
Here's a pull request that adds the ability for Moodle 3.7 custom course fields to be included in the certificate (hasn't had a lot of testing yet, but sharing the work so far here.)

I expect we'll be testing this in the next few weeks - will let you know if I spot any issues.

This is based on the userfields element, and I included course fullname/shortname/idnumber in this as well but feel free to override that and remove fullname from being available via this method!

### DIFF
--- a/element/coursefield/classes/element.php
+++ b/element/coursefield/classes/element.php
@@ -1,0 +1,154 @@
+<?php
+// This file is part of the customcert module for Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


markn86:
Thanks for working on this Dan, very much appreciated. I'll let you guys test it for a bit before integrating.

One thing I would like to point out is the use of `$COURSE` in `render()` and `render_html()` which we should not do - please see https://github.com/markn86/moodle-mod_customcert/issues/82.

markn86:
Actually, not an issue in `render_html` as that is only dummy data used to rearrange the elements.

danmarsden:
thanks for the tip! - I've added a patch to use your helper function in render()

abias:
@danmarsden / @markn86 , just FYI because I just spotted this: This pull request introduced an error in the prechecker results of the plugin (see https://moodle.org/plugins/pluginversion.php?id=19798&smurf=html#phpdoc)

Issue #285, opened 2019-05-22
Add Permissionable hiding of Item in Profile


dustinbrisebois:
Greetings

As per: https://github.com/markn86/moodle-mod_customcert/blob/b211a06bd50559c3e6c1a6d36600307084fa6e5b/lib.php#L370

I am wondering if we can add a permission to hide the "My certificates" item from the user profile list shown here:

![image](https://user-images.githubusercontent.com/16727280/58190660-dcdb0280-7c71-11e9-9500-c8ff3658f4c9.png)

If something like:

`function mod_customcert_myprofile_navigation(core_user\output\myprofile\tree $tree, $user, $iscurrentuser, $course) {
    if (has_capability('mod/customcert:viewallcertificates', $PAGE->cm->context)) {
        $url = new moodle_url('/mod/customcert/my_certificates.php', array('userid' => $user->id));
        $node = new core_user\output\myprofile\node('miscellaneous', 'mycustomcerts',
            get_string('mycertificates', 'customcert'), null, $url);
        $tree->add_node($node);
    }
}`

Instead this would ensure that the permission can remove the list of cerficates.

markn86:
This can be done in CSS. See comment in PR.

dustinbrisebois:
My apologies, I was under the impression that the "viewallcertificates" wasn't for ALL users just all the certificates for your user. 

Doing this in CSS is impractical, it needs to be done with JavaScript AND CSS - which is ugly, compllicated and unnecessary. As there are no proper tags within the box this isn't easy to target/remove from the DOM (as least in the LMS version that I'm using for this).

Is it possible to use the same code I provided but add a new permission that allows for removal from the navigation menu?

PR #286, opened 2019-05-22
#285 - add permission to hide navigation item

dustinbrisebois:


markn86:
Sorry, this is the wrong approach. You are using an unrelated capability - that capability is to check if a user can view any other users certificate on the site. With this change it would mean that students would never be able to see that link, when they are supposed to but they could still access the page if they typed the URL manually in the address bar. A simple CSS change is all that is needed.

PR #294, opened 2019-06-26
Attempt to continue to next certificate/user

mwithheld:
On DB query or other failure, attempt to continue on to the next certificate or user.
This problem blocked sending certificates to other users and sending other certificates.


mdjnelson:
After having a look I can see from your other merge requests that this has been fixed, see #295. Closing.

PR #295, opened 2019-06-26
Do not fail if multiple certificate_issues

mwithheld:
Ignore multiple matches for a user-certificate combo in the certificate_issues table

This means in rare circumstances (i.e. the race condition) the user may get multiple certificates (e.g. one via the UI, one by email) but that is better than not getting one at all and the whole job dying.

### DIFF
--- a/classes/task/email_certificate_task.php
+++ b/classes/task/email_certificate_task.php
@@ -127,7 +127,7 @@ class email_certificate_task extends \core\task\scheduled_task {
 
                 // Ensure the cert hasn't already been issued, e.g via the UI (view.php) - a race condition.
                 $issueid = $DB->get_field('customcert_issues', 'id',
-                    array('userid' => $enroluser->id, 'customcertid' => $customcert->id));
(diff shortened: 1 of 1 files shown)


mwithheld:
I've found 4 more of these in the DB.  The last one is new since my last post.

SELECT userid,customcertid, min(timecreated)
FROM m_customcert_issues
GROUP BY userid, customcertid
HAVING count(userid) > 1
ORDER BY min(timecreated)

1 20426 35 1535058945 Thursday, 23 August 2018 14:15:45 GMT-07:00 DST
2 39740 35 1539646604 Monday, 15 October 2018 16:36:44 GMT-07:00 DST
3 93455 74 1540663516 Saturday, 27 October 2018 11:05:16 GMT-07:00 DST
4 14517 983 1561668669 Thursday, 27 June 2019 13:51:09 GMT-07:00 DST

mdjnelson:
Applied, thanks!

PR #303, opened 2019-08-21
Added element gradefeedback

javierhisyc:
Added element gradefeedback. It allows to show the feedback of a grade element in the certificate.


markn86:
Thanks for working on this!

There is one thing I can think of that this patch does not address, and that is media files left as feedback. See https://tracker.moodle.org/browse/MDL-27520.

mdjnelson:
I am closing this due to the above comments. Please consider placing it at https://moodle.org/plugins/browse.php?list=category&id=73.

PR #304, opened 2019-08-26
Do not fail if multiple certificate_issues fixes #295

mwithheld:
Fix another location of this problem


mdjnelson:
Applied, thanks!

PR #308, opened 2019-09-30
Format text element with format_string() to prevent moodle autolink custom text

diegocardoso93:
In cases when text apresent the same word as course name, moodle is linking that and causing font color change and alignment errors.
This change fix that.


mdjnelson:
The problem with this change is that it will affect users who have HTML in their text fields. I can not merge it as it could potentially break many certificates.

PR #316, opened 2019-12-13
Add userfullname variable for email subject

ak4t0sh:
This PR allow an admin to add userfullname in emails (which can be handy for manager). by using {$a->userfullname} in language pack customization.


mdjnelson:
Done, cheers!

PR #322, opened 2020-02-06
Feature: Consider Course Completion 

dnutzz:
I've added an option for the CustomCert which prints an overview for missing/fulfilled tasks depending on the logic from the core functionality _Course Completion_. There is no custom styling, everything was done with Bootstrap4, which is integrated in the Moodle Core.

## How to use it?
Enable Completion tracking in the course settings and then tick the option _Obtaining certificate requires course to be completed_ in the Customcert activity. In the course settings _Course Completion_ it is possible to determine the conditions when a course is completed (e.g. to obtain a specific course grade, have all activities done and more). All this settings are then reflected in the view from the certificate (see screenshot below).

## Important
This view is based on the core logic of _Course completion_, which works with cronjobs for updating Completion Info. So be sure when using this option to have your cronjobs configured properly. There is a [ticket](https://tracker.moodle.org/browse/MDL-32103) (unfortunately back from 2012) about moving the course completion out of the cron, but I think this won't happen so fast (maybe in Moodle 3.9 or 4.0?).

image

### DIFF
--- a/classes/certificate.php
+++ b/classes/certificate.php
@@ -479,128 +479,125 @@ class certificate {
      * @return bool true if completed, otherwise false
      */
     public static function get_completion_requirements($course) {
-        global $USER, $PAGE, $COMPLETION_CRITERIA_TYPES;
+        global $USER, $PAGE;
         $info = new \completion_info($course);
         $completions = $info->get_completions($USER->id);
         $modinfo = get_fast_modinfo($course);
 
-        // Prepare return object
-        $completion_view = new stdClass();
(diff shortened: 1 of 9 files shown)


dnutzz:
Will work on the Travis CI build fail and let you know when it's finished.

mdjnelson:
Hi there, sorry for getting to this so late - I was not working with Moodle for some time while I was away in Barcelona. Are you able to post about this on the forums to see if there would be popularity in such a feature? See https://moodle.org/plugins/browse.php?list=category&id=73.

PR #326, opened 2020-02-22
Fix of showing names of custom user fields

serg123e:
Show normal human-readable name of custom user profile fields instead of internal shortname


mdjnelson:
Thanks! Cherry-picked.

PR #338, opened 2020-03-31
Do not fail if multiple certificate issues (#295)

Peterburnett:
NOTE: This is a cherry-pick of another commit. Original message below.

Ignore multiple matches for a user-customcert combo in the customcert_issues table.

This means in rare circumstances (i.e. the race condition) the user may get multiple certificates (e.g. one via the UI, one by email) but that is better than not getting one at all and the whole job dying.


mdjnelson:
Thanks Peter,

3.5 is no longer supported but since this is being used by Catalyst i'll make an exception ;)

Cheers.

PR #343, opened 2020-04-24
Fix get position when browser shrinking

JBThong:
When we go to 'Reposition elements' to display, and shrinking the browsers, the image is overlapped the button when shrinking the page. So we need to check the browser resizes and adjust position of image.
I have attached issue image and result in below:
issue
result


mdjnelson:
Sorry it took me so long to get to this. I no longer work at Moodle HQ and am focusing on other projects.

mdjnelson:
Thanks for this. Just testing it out now but seems to be working very well.

mdjnelson:
Done. :) Thanks!

rezaies:
All elements go back to their original place after a window resize. Is this expected?

mdjnelson:
It makes sense to me. Why do you think they shouldn't? It is how a browser behaves. However, open to discussion. :)

rezaies:
I don't understand "It is how a browser behaves".
let me write reproduction steps:
1. Create a certificate
1. Add 3 text elements (place them at 0,0) to it
1. Go to the Drag&drop design page
1. Position the elements as you wish. put the first one in the center, the second one a bit above it, and third one a bit below it.
1. Maximise the page to see the page better
Bingo, all elements moved back to their original place at 0,0. Do you think this really is how it is supposed to be?

mdjnelson:
I didn't know what you meant by original position but yes this behaviour is obviously wrong. Either I revert this or we fix it in a separate issue.

PR #351, opened 2020-06-05
Added course average grade

abertranb:
Hi! we added an element to allow select some courses to calculate the average grade.

Best regards


mdjnelson:
Have you checked on the forums whether or not this would be in much use? There is a section to store plugins at https://moodle.org/plugins/browse.php?list=category&id=73 where you could upload this and then create a forum post seeing if it should be merged or not.

PR #362, opened 2020-07-05
Fix custom fields not displaying properly

StevenTsvetkov:
Hey Mark,

This patch fixes the custom fields. It turns out that date and checkbox fields also didn't show up in a human-readable format, this fixes them as well.

Cheers! 🙂

### DIFF
--- a/element/coursefield/classes/element.php
+++ b/element/coursefield/classes/element.php
@@ -121,7 +121,7 @@ class element extends \mod_customcert\element {
     }
 
     /**
-     * Helper function that returns the text.
+     * Helper function that returns the field value in a human-readable format.
      *
      * @param \stdClass $user the user we are rendering this for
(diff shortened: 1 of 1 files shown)


mdjnelson:
Looks good, cheers! Do you know if this should be applied to earlier branches. I won't close for now in-case that is needed.

StevenTsvetkov:
I've tested this on 3.8 and 3.7 with their respective branches, all had this issue and it was resolved with this patch. The 3.8 branch also works as expected on 3.9.

I can check 3.6 and 3.5 if you want, do you want me to open PRs for those or are you going to merge the changes yourself?

mdjnelson:
I am happy to cherry-pick this to past branches myself. If you could confirm 3.6 and 3.5 that would be a big help and save me some time.

PR #371, opened 2020-08-23
Support right to left certificates

rezaies:


mdjnelson:
Excuse my ignorance but can't people using RTL languages just move the words to the right instead of the left?

rezaies:
No. What they see on the drag and drop designer page is not what the actual
pdf will look like. They put an element on the left of the designer page
and to their surprise they see that element on the right on the pdf.

rezaies:
Also direction and alignment are 2 separate things

rezaies:
Hi Mark,
I was wondering if this is still under your radar or probably forgotten :)

mdjnelson:
Definitely has not been forgotten. It is in the list of issues I am going to do before releasing a new version. I am in the middle of moving countries and right now am working on other things to make much needed money rather than my open source projects. This should settle down once I am back in Australia.

mdjnelson:
So, I have worked with RTL languages before. In this case I still don't see why a user can not type out the string in a RTL language and then position it accordingly. I would download the Hebrew language pack and check how this works when filling in an input field but the internet is so slow at the hotel I am being quarantined at that it fails to download. What happens on input fields, do the new characters get added to the start of the input field when you are typing? Can you please provide an example for my ignorant mind on what this setting does. Pretend you are writing testing instructions on Jira :) Also, can you address my PR comments, an ignorant person like myself will have no idea what direction means when editing the certificate. Thanks Shamim!

rezaies:
Hi Mark,
First of all, an RTL designed certificate should always remain RTL no matter what language its designer use and an LTR certificate should also remain LTR no matter what language its designer use. Being LTR or RTL is an attribute of the certificate, and should be independent of the teacher's language.

Apart from the above, here is just a case that the current code fails.
### Fail case 1
1. Log in as an admin
1. Switch your language to an RTL language
1. Go into a course and create a certificate activity
1. Add a text element to the certificate and position the text at roughly 2 cm from the top of the page and 1 cm from the RIGHT of the page
1. Save your design
1. Click on save changes and preview
1. Download the PDF somewhere and then open it
1. You expect the text to be at 2cm from the top and 1cm from right, but you see that the text is 2cm from the top and 1cm from left!

amd/src/rearrange-area.js
```diff
@@ -117,8 +117,13 @@ define(['jquery', 'core/yui', 'core/fragment', 'mod_customcert/dialogue', 'core/
 
             RearrangeArea.prototype._setPosition = function(elementid, refpoint, posx, posy) {
                 var element = Y.one('#element-' + elementid);
+                var isRtl = Y.one('#pdf').getComputedStyle('direction') === 'rtl';
 
-                posx = Y.one('#pdf').getX() + posx * this.PIXELSINMM;
+                if (isRtl) {
+                    posx = Y.one('#pdf').getX() + parseFloat(Y.one('#pdf').getComputedStyle('width')) - posx * this.PIXELSINMM;
+                } else {
+                    posx = Y.one('#pdf').getX() + posx * this.PIXELSINMM;
```
rezaies, line 125: can anyone else see this?

mdjnelson:
Closing. I have many people using this in RTL languages with no issues.

PR #372, opened 2020-08-24
send a copy of a certificate to a certain (student dependant) receiver

timopitt:
As we needed an ability to send a copy of a certificate to a certain (student dependant) receiver i implemented some functions to select a custom field from the user profile to send the certificate to the mail address in that field.

Now you can select if the certificate will be sent to the student, to the teacher, to one (or more) static email(s) and/or to a mail address in an custom user field.

To store the additional information i added an column in the database and the corrensponding methods for updating the module.


mdjnelson:
Thanks for working on this.

1. Please remove all comments that say ''added by Timo Pitt". If we did this for Moodle core it would make reading the code very tedious.
2. You have a install.newxml file, this should just be an edit of the install.xml file.
3. Strings have to be ordered alphabetically by their index.
4. What is the point of the site setting? Isn't a module site enough?
5. Version should only be bumped by 1 on the 3.8 branch.
6. This will confuse users who do not have access to the user profile fields and are not sure what each one represents, it's also impossible for a Date/time profile field to be an email so displaying these is useless.

I can kind of understand what you are trying to do here but to me this seems a bit too hacky and is abusing the custom profile fields. Can you please give me a scenario where this is necessary? Like, if the user has a tutor you could email the user and their tutor by putting their tutors email as a custom profile field?

PR #373, opened 2020-08-25
Add new element to display course short name

emyb2:
This element is based on the coursename element, and provides the ability to display the course shortname on certificates

### DIFF
--- a/element/courseshortname/classes/element.php
+++ b/element/courseshortname/classes/element.php
@@ -1,0 +1,73 @@
+<?php
+// This file is part of the customcert module for Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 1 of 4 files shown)


emyb2:
I hope that this is a pretty straight forward PR as I would very much appreciate this to be added to the main activity module :)

mdjnelson:
Thanks Donald. I have added it to my to-do list (keep in mind I no longer work at Moodle HQ).

emyb2:
Thanks Mark. I will await the pulling of this pull request :)

mdjnelson:
Just thinking .. instead of adding another element, perhaps we could edit the existing element to have a select box where you can choose the full name or the short name. Thoughts?

emyb2:
Sounds good. That would be dryer.

mdjnelson:
Would you be willing to code it? :)

emyb2:
Haha was waiting for this. I can have a crack at it ;) any existing element you have in mind I can use to base it off?

mdjnelson:
The grade element has a field to choose what grade item to choose, that's a good start. Cheers!

mdjnelson:
I created #415, closing.

Issue #121, opened 2017-07-19
Add ability to set alignment for all text elements


rlowry:
It would be very useful to be able to set the alignment (left, right, center, justified) for all elements that output text e.g. the text element, the course name element and the category name element etc.

Center alignment of text can currently be achieved by setting the elements **Position X** to half the page width and setting the **Reference point location** to Center. This works fine when the text does not wrap e.g.

![example1](https://user-images.githubusercontent.com/7382210/28361708-8cdf1ec2-6c71-11e7-8e50-8bec266a418f.png)

However when the text wraps due to size / length the text is output left aligned e.g.

![example2](https://user-images.githubusercontent.com/7382210/28361762-b06a7dfa-6c71-11e7-8529-fe5881fe8e96.png)

tracswa:
Came here to say above, agree it would very useful! It's particularly important when using a field like student's name and you don't know exactly how many lines or how wide the text will be.

smlavoie:
I'd rather have the option to align any text element left, center, right or justify (instead of all centered).

PR #386, opened 2020-10-27
Add support for the horizontal alignment of elements

Kencho:
This PR allows to specify the horizontal alignment of elements on a per-element basis, as requested in the issue #121 

Note this applies only to the alignment of the contents of the elements. The element itself still needs to be manually placed in the page. Also, whether the alignment will occur in the end or not depends on the element type itself; some elements may not support horizontal alignment of their contents.

Adds a new field `alignment` in the `customcert_elements` table, defaulting to `'L'`. This can be edited in the element edition form.

No version numbers were changed, although the addition is clearly separated in `db/upgrade.php` so it's just a matter of using the final version number there.

There are still slight discrepancies between the arrangement editor and the final rendering. This is because the rendering collapses the elements' width and then does the element alignment based on this width and the element anchor (reference) point, whereas the arrangement editor uses the configured element width. Thus, small elements (e.g., non-wrapping texts) are rendered centered although the editor displays it using the full width. This is beyond the scope of this PR.

### DIFF
--- a/classes/element.php
+++ b/classes/element.php
@@ -38,6 +38,21 @@ defined('MOODLE_INTERNAL') || die();
 abstract class element {
 
     /**
+     * @var string The left alignment constant.
+     */
+    const ALIGN_LEFT = 'L';
+
+    /**
+     * @var string The centered alignment constant.
+     */
+    const ALIGN_CENTER = 'C';
+
+    /**
+     * @var string The right alignment constant.
+     */
+    const ALIGN_RIGHT = 'R';
+    
+    /**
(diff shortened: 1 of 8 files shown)


mdjnelson:
Sorry for delay in reviewing this. I am quite busy right now with personal issues but I hope that I can look eventually.

Kencho:
No problem :) Life is strange these days, and besides, there's no hurry. Hope everything is fine.

marcellobarile:
news about the PR?

mdjnelson:
Looking at it this week all going well.

mdjnelson:
Doing it now!

PR #389, opened 2020-11-10
Adding relative expiry date options to certificates.

LegacyXSoftware:
I have added five options for relative expiry dates.  Each option adds an x number of years to the issued date, allowing for certificates to expire after a period of time.  The options included are: 1 year, 2 years, 3 years, 4 years and 5 years.


mdjnelson:
Thanks. I'll get round to this when I do my next release.

Note - I just did a release yesterday so there may be some time until I deal with this. :)

mdjnelson:
Can I have some context as to why this might be used?

mdjnelson:
Sorry for taking so long to look at this. I have a lot of issues to go through and am only one person. :(

alihamas575:
So it would be a nice feature for me. Because there are legally certain periods in the trainings we provide, for example, a subject related to employee health must be repeated by some employees every year, for others every 3 years.

LegacyXSoftware:
Apologies for the very late reply.  As @alihamas575 has said many certificates (safety, health, etc.) require expiry dates and this is what we have implemented to provide clients with an easy way to add expiry dates.

mdjnelson:
Why not use a text element and just set the date in advanced but written?

alihamas575:
> Why not use a text element and just set the date in advanced but written?

Greetings, first of all, thank you for your response, I am not opening related courses at the same time. I have 3 courses for example, A, B, C courses

My courses are always open and new people are signing up every day. For example, Micheal user registered for course A today and finished the training 10 days later. The certificate issuance date will be 18.06.2021 and this certificate (the course I have given has a legal validity period) will be valid for 1 year.

User Gabriel registered for course B after 6 months. After 5 days, he finished the course, the certificate issuance date will be 13.12.2021 and this person's certificate will be valid for 2 years.

etc. When I write as text for reasons like this, it doesn't change much if I write it as text because the course completion date and duration are different for everyone, but if I can run a formula, it seems like this can be fixed.

Do you have any suggestion about it?
For example, if the person completed the course in course A today, if I add a formula, add +365 days to the course end date in the text field, and write the certificate expiry date, at least in this way, I would have produced a temporary solution.

LegacyXSoftware:
@alihamas575 explains its use case very well,  I have nothing else to add!

mdjnelson:
Ok, thanks all. I'll add this to the next release.

mdjnelson:
Cherry-picked this just then. Looking at pushing it forward if all is well.

alihamas575:
Has improvement been made?
So this means it's ok?
Can we now set a deadline when we add a certificate date?

@mdjnelson

mdjnelson:
I need to test this but if all goes well then yes it will up on github soon.

alihamas575:
> I need to test this but if all goes well then yes it will up on github soon.

Thank you so much. We eagerly await.

alihamas575:
I'll post it in the relevant thread. But by the way, I have to ask here. To check the certificates, it will be a better validation if it brings up the whole certificate when you type the certificate number in the verification field.

Because in the current situation, when we type the barcode or number, it only says that the certificate is valid or not valid. This is an easy way to fool people. When I find the certificate number from anywhere, we can change the certificate owner's name and surname and show the same certificate code as if it is correct for hundreds of people.

mdjnelson:
Closing as I added this to all supported branches.

mdjnelson:
I only added this to 4.0 and 3.11. Thinking if I should backport to 3.10 and 3.9 since it's technically an improvement ..

mdjnelson:
@Dyc9l you add a date element and select how many years it should expire on.

mdjnelson:
Closing. I am going to rewrite this to make it so you can specify an exact date.

PR #390, opened 2020-11-18
Add username to userfield form elements

uvigo-atic:
Allows choosing the username as a userfield form element


mdjnelson:
Done! Thanks!

Issue #412, opened 2021-03-18
teachers/manager cannot open issued certificates without mod/customcert:receiveissue 


Urpokarhu1:
Hello,

teacher/manager-role cannot open certificates that are issued to students if you don't allow mod/customcert:receiveissue if  this isn't allowed issued certificate opens up as blank white page for them.

If you enable this permission they will receive email message everytime when someone creates a certificate with setting "Email students" "yes".

This need to be fixed. @mdjnelson

stonerl:
Should be solved by #413

mdjnelson:
Added #413 to do as next release.

mdjnelson:
Fixed. :)

PR #413, opened 2021-03-19
alllow manager to download certificates

stonerl:
Currently, it is not possible for editing teachers and managers to download certificates for users because we only check for $canreceive.


mdjnelson:
Cherry-picked, thanks.

mdjnelson:
Commit message contains https://github.com/mdjnelson/moodle-mod_customcert/pull/412.

Issue #420, opened 2021-04-15
PDF is not printing when debugging is ON


golenkovm:
Hello,

I've just found that PDF is not displaying when debuging is ON. Debugging messages seem to be a reason:

![image](https://user-images.githubusercontent.com/22066290/114815652-fe12e200-9df9-11eb-9192-d8c280c0286f.png)

golenkovm:
Root notice:

```
moodle_1   | 2021/04/15 04:56:03 [error] 18#18: *14120 FastCGI sent in stderr: "PHP message: PHP Notice:  Trying to get property 'protection' of non-object in /siteroot/mod/customcert/classes/template.php on line 275PHP message: PHP Stack trace:PHP message: PHP   1. {main}() /siteroot/mod/customcert/edit.php:0PHP message: PHP   2. mod_customcert\template->generate_pdf() /siteroot/mod/customcert/edit.php:233PHP message: PHP Notice:  Trying to get property 'deliveryoption' of non-object in /siteroot/mod/customcert/classes/template.php on line 281PHP message: PHP Stack trace:PHP message: PHP   1. {main}() /siteroot/mod/customcert/edit.php:0PHP message: PHP   2. mod_customcert\template->generate_pdf() /siteroot/mod/customcert/edit.php:233PHP message: Default exception handler: Exception - TCPDF ERROR: Some data has already been output, can't send PDF file Debug: 
moodle_1   | Error code: generalexceptionmessage
moodle_1   | * line 2950 of /lib/tcpdf/tcpdf.php: Exception thrown
moodle_1   | * line 7642 of /lib/tcpdf/tcpdf.php: call to TCPDF->Error()
moodle_1   | * line 190 of /lib/pdflib.php: call to TCPDF->Output()
moodle_1   | * line 329 of /mod/customcert/classes/template.php: call to pdf->Output()
```

golenkovm:
Protection and delivery options seem to be a regression of https://github.com/mdjnelson/moodle-mod_customcert/commit/8446a2e10ac7f662438b47d1f4b6e2bad4c6a1d4
while moving `$PAGE` callsjust a general fix

mdjnelson:
Fixed for 3.10, 3.9 and 3.8. Thanks.

PR #421, opened 2021-04-15
Fix the issue with displaying PDF when debugging is ON

golenkovm:
This is a fix for #420


mdjnelson:
Cherry-picked. Thanks.

PR #422, opened 2021-04-15
Fix the issue with displaying PDF when debugging is ON

golenkovm:
This is a fix for #420 for Moodle 3.9 branch


mdjnelson:
Will backport https://github.com/mdjnelson/moodle-mod_customcert/pull/421 which has more changes (`classes/template.php`).

Issue #433, opened 2021-06-29
Multi-language issues


michael-milette:
There are a few places in the plugin that are not reflecting the current UI language in Moodle. In my case, I am using the Multi-Language Content v2 filter which uses plain-text {mlang} tags.

When viewing or previewing the PDF, the page title is not processed through output filters (screen shot from Chrome). This is the same string based on the HTML page title tag which also controls the content of the browser tab.
![image](https://user-images.githubusercontent.com/3808579/123871568-c77d3b80-d901-11eb-8c4d-e48a87c23a09.png)

When viewing the list of created templates, the template names are not processed through output filters.
![image](https://user-images.githubusercontent.com/3808579/123871814-322e7700-d902-11eb-9e4f-17c00d569171.png)

When creating/editing a certificate, the list of available templates is not processed through output filters.
![image](https://user-images.githubusercontent.com/3808579/123872050-820d3e00-d902-11eb-999b-e61246d8d109.png)

Pull requests to fix Custom Certificate for Moodle 3.8, 3.9, 3.10 and 3.11 will be coming shortly.

Best regards,

Michael Milette

mdjnelson:
Thanks, will have a look soon and will ship with next release. Note - I no longer support 3.8.

mdjnelson:
Thanks for this work Michael. :) Integrated in 3.9, 3.10 and 3.11.

PR #435, opened 2021-06-29
Fix for multi-language issues.

michael-milette:
- Filename of PDF when viewing/previewing PDF.
- Page title tag when viewing/previewing PDF.
- List of available templates.
- Template Load dropdown list.


michael-milette:
This is in reference to issue #433.

mdjnelson:
Integrating now. :)

PR #451, opened 2021-08-06
Closes #449 The user id should be mapped to the equivalent user id in new sites during activity restore

### LINKED ISSUE #449: Backup and restore does not store who has already received the certificate email
There might be something related with this issue:

https://github.com/mdjnelson/moodle-mod_customcert/issues/379

andrewmadden:
As the userid in the customcert_issues table were using the userid from the site where the activity was backed up, emails were being sent to those that had already received them. Theoretically there wwere also users who should have received an email who didn't get one.


mdjnelson:
Thanks Andrew but 3.8 is out of support, closing.

Issue #469, opened 2021-10-25
Setting: View a custom certificate mod/customcert:view to "allow" exposes users across all cohorts


Helicopter9005:
Hello,

Firstly thanks for the great customer certificate plugin - it is awesome!

Our goal is to setup different cohorts and teachers for each cohort that need to monitor the progress and certificates for each student that belongs to the same cohort like the teacher.

We've set up the cohorts and groups, so that the users don't see each other.

When we set "View a custom certificate mod/customcert:view" to "allow" it exposes users across all cohorts. 

We do the setup in the following way:
1. Setup a new role with the ARCHETYPE: Teacher (non-editing)
2. Change the activity:
    View a custom certificatemod/customcert:view  to "Allow"

Then login as a user with the role you just setup and:
1. Go to any course you are enrolled in and click on the participants overview ( my.moodle.website/course/view.php?id=28 )

![image](https://user-images.githubusercontent.com/55496293/139957871-510bb5d7-28bc-40c2-a137-934d9f320792.png)

2. Here the teacher only sees the students of their cohort - which is fine. ( my.moodle.website/user/index.php?id=28 )
![image](https://user-images.githubusercontent.com/55496293/140038430-691b9cdf-25b6-4793-bb30-1ec23b45485d.png)

3. If the teacher clicks someone on the participants list - it will take them to the profile page of the user (  my.moodle.website/user/view.php?id=522&course=28 )
4. Click on "My certificates" under the "Miscellaneous" section

![image](https://user-images.githubusercontent.com/55496293/139958613-ec2a3c84-3771-42b4-b562-ea2a7a00d1eb.png)

5. Then click on Dashboard > Users at the top of the page ( my.moodle.website/mod/customcert/my_certificates.php?userid=522 )

![image](https://user-images.githubusercontent.com/55496293/139959006-9abf8ebc-71e2-42ae-8778-ef517b96108f.png)

6.  You are taken to a page where you then see all the users in the system. It is not limited by cohort, course, group. ( my.moodle.website/user/index.php?id=1 )
![image](https://user-images.githubusercontent.com/55496293/139959639-eea70f10-9fbb-4b8d-ad27-cc72d86352b9.png)

How can we stop this from happening?

Thanks!

mdjnelson:
Cherry-picked. Thanks @brendanheywood.

PR #470, opened 2021-11-04
Add roundtrip linking between my profiles certs and courses #469

brendanheywood:
#469

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -370,7 +370,13 @@ function customcert_extend_settings_navigation(settings_navigation $settings, na
  * @return bool
  */
 function mod_customcert_myprofile_navigation(core_user\output\myprofile\tree $tree, $user, $iscurrentuser, $course) {
-    $url = new moodle_url('/mod/customcert/my_certificates.php', array('userid' => $user->id));
+    $params = [
+        'userid' => $user->id
+    ];
+    if ($course) {
+        $params['course'] = $course->id;
+    }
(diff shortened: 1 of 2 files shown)


mdjnelson:
Hi @brendanheywood - I can't see what this achieves except passing the `$courseid` to `require_login`. There is no change in navigation afaict.

brendanheywood:
Passing the $courseid to require_login changes the way core behaves and it adds the course context to the crumb trail and other navigation elements. So from a course you click into your plugin, and then you click into various core navigation and get back to where you came from. Without it you end up at the site level instead of where you started.

Eg:

1. from /user/index.php?id=2
2. click on a user -> /user/view.php?id=105&course=2
3. click on participants, and you are back to /user/index.php?id=2
4. click on user again -> /user/view.php?id=105&course=2
5. now click on 'my certificates' -> /mod/customcert/my_certificates.php?userid=105&course=2
6. now you are still 'inside' the course, so in the crumb trail you can click on either
7. /user/index.php?id=2 - back to start
8. /course/view.php?id=2 back to course

brendanheywood:
Compare with:

image

vs without:

image

mdjnelson:
I thought this was the case but it does not work for me. The breadcrumbs do not change. It remains "Dashboard > Profile > My certificates".

mdjnelson:
Ahhh I see! Sorry. I was doing it as the user, not the administrator. Cherry-picking.

PR #472, opened 2021-11-17
Add field exist check for alignment field

aspark21:
Upgrading from MOODLE_310_STABLE to MOODLE_311_STABLE fails as field already exists


mdjnelson:
Thanks, pushed.

PR #490, opened 2022-04-06
Custom email feature added

btc989:
We had a request from one of our user groups to customize the emails sent to students. Instead of just changing the lang string we wanted to develop a feature that would allow us to change the emails dynamically, so that we would not have to wait for an upgrade to occur in order to purge the caches. We thought others might be interested in this feature, hence the pull request. Some further adaptions to allow a test email to be sent will be worked on if time permits.


andrewhancox:
I've got another client with a similar requirement - this is a really good enhancement.

Issue #512, opened 2022-08-20
Error thrown when we have custom profile fields


codigo-iot:
This problem has reappeared in version 4.0+

_Originally posted by @codigo-iot in https://github.com/mdjnelson/moodle-mod_customcert/issues/465#issuecomment-1221224353_

Información de depuración: Unknown column 'uf1d_1.data' in 'field list'
SELECT u.id, u.picture, u.firstname, u.lastname, u.firstnamephonetic, u.lastnamephonetic, u.middlename, u.alternatename, u.imagealt, u.email, u.username, u.idnumber, u.phone1, u.phone2, u.department, u.institution, u.city, u.country, uf1d_1.data AS profile_field_telefono, ci.id as issueid, ci.code, ci.timecreated
FROM mdl_user u
INNER JOIN mdl_customcert_issues ci
ON u.id = ci.userid
WHERE u.deleted = 0
AND ci.customcertid = ?
AND NOT u.id IN (?,?,?,?,?,?,?,?,?,?)
ORDER BY CONCAT(firstname, ' ', lastname) LIMIT 0, 50
-- line 291 of /mod/customcert/classes/certificate.php: call to mysqli_native_moodle_database->get_records_sql()
-- line 204 of /mod/customcert/classes/report_table.php: call to mod_customcert\certificate::get_issues()
-- line 2132 of /lib/tablelib.php: call to mod_customcert\report_table->query_db()
-- line 158 of /mod/customcert/view.php: call to table_sql->out()
[array (
0 => '7',
1 => 3388,
2 => 128,
3 => 3515,
4 => 3509,
5 => 3513,
6 => 275,
7 => 2,
8 => 275,
9 => 320,
10 => 3522,
)]
Error code: dmlreadexception

i-am-sheep:
I've had the same issue with version 4.0.2 in moodle 4

Debug info: Unknown column 'uf1d_1.data' in 'field list'
SELECT u.id, u.picture, u.firstname, u.lastname, u.firstnamephonetic, u.lastnamephonetic, u.middlename, u.alternatename, u.imagealt, u.email, u.username, uf1d_1.data AS profile_field_attendancedate, ci.id as issueid, ci.code, ci.timecreated
FROM mdl_user u
INNER JOIN mdl_customcert_issues ci
ON u.id = ci.userid
WHERE u.deleted = 0
AND ci.customcertid = ?
AND NOT u.id IN (?,?) 
[...]

Issue #465, opened 2021-10-14
Error thrown when we have custom profile fields for Moodle 3.11


vuvanhieu143:
Hi @mdjnelson ,
I found out there will be an error when we have custom fields in user profile
This code should include the join and params for custom fields.
` $ufields = \core_user\fields::for_userpic()->including(...$extrafields);
      $ufields = $ufields->get_sql('u', false, '', '', false)->selects;
      $sql = "SELECT $ufields, ci.id as issueid, ci.code, ci.timecreated
                FROM {user} u
          INNER JOIN {customcert_issues} ci
                  ON u.id = ci.userid
               WHERE u.deleted = 0
                 AND ci.customcertid = :customcertid
                     $conditionssql";
      if ($sort) {
          $sql .= "ORDER BY " . $sort;
      } else {
          $sql .= "ORDER BY " . $DB->sql_fullname();
      }`

These are the code changes that will fix this issue.

 $ufields = \core_user\fields::for_userpic()->including(...$extrafields);
        [
                'selects' => $userfieldsselects,
                'joins' => $userfieldsjoin,
                'params' => $userfieldsparams
        ] = (array) $ufields->get_sql('u', true);

        $allparams = array_merge($allparams, $userfieldsparams);
        $sql = "SELECT ci.id as issueid, ci.code, ci.timecreated $userfieldsselects
                  FROM {user} u
            INNER JOIN {customcert_issues} ci
                    ON u.id = ci.userid
                    $userfieldsjoin
                 WHERE u.deleted = 0
                   AND ci.customcertid = :customcertid
                       $conditionssql";

Sorry, I can't submit a PR right now because I'm busy with my work.
I will submit a PR later.

Could you please take a look and give me your thought on this issue.

mdjnelson:
Thanks @vuvanhieu143. Looks like it works. This has been mentioned in the forums and in another issue but I could not replicate it so I appreciate you coming up with a solution and a PR.

vuvanhieu143:
Thanks, @mdjnelson ,I will submit a PR and behat scenario for this case later.

vuvanhieu143:
Hi @mdjnelson , I have submit a PR
#466

codigo-iot:
This problem has reappeared in version 4.0+

PR #517, opened 2022-10-14
feat: fix: Error thrown when we have custom profile fields

kelsoncm:
Hi, @mdjnelson !

This error occurs because when there a mandatory field in the users profile, the JOINs with the 'user custom field' table's must be included. As Moodle already has the solution ready for this, in this patch I used the native solution..

I hope it helps.

See: https://github.com/mdjnelson/moodle-mod_customcert/issues/512

### DIFF
--- a/classes/certificate.php
+++ b/classes/certificate.php
@@ -266,32 +266,22 @@ class certificate {
             return array();
         }
 
-        // Add the conditional SQL and the customcertid to form all used parameters.
-        $allparams = $conditionsparams + array('customcertid' => $customcertid);
-
         // Return the issues.
         $context = \context_module::instance($cm->id);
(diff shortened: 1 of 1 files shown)


atexeira (approved):
This update solve the problem.

mdjnelson:
Thanks for your work on this. Will merge when I am at my computer.

mdjnelson:
Computer wasn't playing nice but all sorted now with new battery (I hope). Will get this done tonight (Perth time).

mdjnelson:
Sorry I am not sure why I closed this. Getting confused from all the similar issues.

mdjnelson:
I pushed this commit to 4.0. However, this was already fixed by https://github.com/mdjnelson/moodle-mod_customcert/issues/465 but this SQL is nicer which is why I am using it now.

PR #522, opened 2022-11-05
521 Fix: title and description shown twice

leonstr:
Removed code to show this activity's title and description as this is handled by the activity header class in Moodle 4.0 (see MDL-72413 and https://moodledev.io/docs/devupdate#the-activity-header-class).

This is a new MOODLE_400_BRANCH with $plugin->requires = 2022041900 because this change is specifically for Moodle 4.0 or later.

### DIFF
--- a/version.php
+++ b/version.php
@@ -24,10 +24,10 @@
 
 defined('MOODLE_INTERNAL') || die('Direct access to this script is forbidden.');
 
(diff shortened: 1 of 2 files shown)


leonstr:
This change needs more work as there are other affected pages in the plugin.

PR #523, opened 2022-11-06
521 Fix: title and description shown twice

leonstr:
Removed code to show this activity's title and description as this is now handled by the activity header class in Moodle 4.0 (see MDL-72413 and https://moodledev.io/docs/devupdate#the-activity-header-class).

Propose creating branch MOODLE_400_STABLE as these are Moodle 4.0-specific changes. Bumped version number for the same reason.


mdjnelson:
I will create a `MOODLE_400_STABLE` branch tonight and cherry-pick these commits. Thanks.

PR #530, opened 2023-01-03
Do not write unnecessary lines to issue table

tobbetto:
Added condition to ignore writing to table issues when has_capability is above student

### DIFF
--- a/view.php
+++ b/view.php
@@ -157,7 +157,13 @@ if (!$downloadown && !$downloadissue) {
     if ($downloadown) {
         // Create new customcert issue record if one does not already exist.
         if (!$DB->record_exists('customcert_issues', array('userid' => $USER->id, 'customcertid' => $customcert->id))) {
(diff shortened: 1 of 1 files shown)


view.php
```diff
@@ -157,7 +157,13 @@
     if ($downloadown) {
         // Create new customcert issue record if one does not already exist.
         if (!$DB->record_exists('customcert_issues', array('userid' => $USER->id, 'customcertid' => $customcert->id))) {
-            \mod_customcert\certificate::issue_certificate($customcert->id, $USER->id);
+            /**
+             * Do not write a line in the table if the user is admin, teacher or manager and only looking at a student's certificate.
+             * Only students viewing their certificate are written as a row in the table, otherwise do nothing.
+             */
+            if(!has_capability('enrol/manual:enrol', $context)){
```
mdjnelson, line 164: I swear I commented on this already but can't see where it has gone. Just to repeat myself we can't use the capability `enrol/manual:enrol` as people can be enrolled in a course via numerous ways.

Issue #501, opened 2022-05-30
Coding problem: unsupported modification of PAGE->context from 70 to 70 ++


ewallah:
Using the global $PAGE in a scheduled task only works when you execute the task inside the browser. During cron, there is no $PAGE available.

Execute scheduled task: mod_customcert\task\email_certificate_task
... started 05:21:02. Current memory use 29MB.
++ Coding problem: unsupported modification of PAGE->context from 70 to 70 ++
* line 1064 of /lib/pagelib.php: call to debugging()
* line 90 of /mod/customcert/classes/task/email_certificate_task.php: call to moodle_page->set_context()
* line 253 of /lib/cronlib.php: call to mod_customcert\task\email_certificate_task->execute()
* line 120 of /lib/cronlib.php: call to cron_run_inner_scheduled_task()
* line 73 of /lib/cronlib.php: call to cron_run_scheduled_tasks()
* line 178 of /admin/cli/cron.php: call to cron_run()
... used 802 dbqueries
... used 0.46844792366028 seconds
Scheduled task complete: mod_customcert\task\email_certificate_task

ewallah:
A possible solution can be to use a $page variable instead of the global $PAGE [see](https://github.com/mdjnelson/moodle-mod_customcert/commit/17851b70e309a5f162d9b4818334584e8535147e)

This way you avoid using the global $PAGE, as there is no global $PAGE available during cron, the function runs in the background without user interface.

TopSolid:
> A possible solution can be to use a $page variable instead of the global $PAGE [see](https://github.com/mdjnelson/moodle-mod_customcert/commit/17851b70e309a5f162d9b4818334584e8535147e)
> 
> This way you avoid using the global $PAGE, as there is no global $PAGE available during cron, the function runs in the background without user interface.
Thank you @ewallah, it worked on my instance!

mdjnelson:
Closing as a duplicate of #443. @ewallah are you able to create a pull request to fix this?

wjroes:
The error is " Coding problem: unsupported modification of PAGE->context from 70 to 70" and the problem is trying to change the context on the global $PAGE variable, but since it tries to change to the same value why don't you just leave that statement $PAGE->set_context($context); out?

PR #541, opened 2023-03-13
#501 unsupported modification of PAGE->context from 70 to 70

ewallah:


ewallah:
Refused to fix after more than 6 months.

Issue #518, opened 2022-10-17
Trigger event when certificate is changed


leonstr:
If someone edits a certificate layout or edits an element I think there should be an event triggered resulting in an entry in the Moodle logs.

For example:
**Time:** 17 Oct 2022 15:24, **Component:** Custom certificate, **Description:** The user with id '2' edited the certificate of 'customcert' activity with course module id '485'.

mdjnelson:
Feel free to create a MR and I will review it. :)

mdjnelson:
Merged. Thanks.

leonstr:
Thanks Mark. Apologies for the work needed on top of the PR.

mdjnelson:
That's OK @leonstr. I got ahead of myself thinking it would be quick to add a bunch more of events and unit tests but no, not a small task at all. 😄

If you could however look at the changes I made and give me a +1 that would be great since no one reviewed them.

leonstr:
Hi @mdjnelson there are three issues I can see:

1. Creating a new certificate or template results in both a `template_created` and a `template_updated` event. The latter is a result of `add_page()` triggering both a `page_created` and a `template_updated`. Having both `template_created` and `template_updated` on creating a template doesn't make sense to me as 'create' is one action.
2. Adding a new element results in two `template_updated` events. The second one is a result of `save_page()` triggering both a `page_updated` and a `template_updated`. Again, two events -- this time the same event -- for one action feels like a bug.
3. If the certificate has more than one page and you add an element to one page then a `page_updated` is triggered for each page even though only one has been changed.
4. For the `element_*` and `page_*` events I think `get_description()` should use 'the', not 'a'/'an', e.g. _The user with id '...' updated the page with id '...'_, not _The user with id '...' updated the page with id '...'_. Trivial but the later reads wrong to me, and it seems to be what Moodle uses, e.g. for course events.

I'll try to find time to see if I can patch these in which case I'll open an issue with a PR for you to review. For 1. and 2. it's not a case of simply removing the `template_created` and `template_updated` from `add_page()` and `save_page()`, otherwise there'd be no `template_updated` when you click **+ Add page**.

mdjnelson:
Thanks a lot @leonstr, it would be much appreciated if you could create a MR for this.

mdjnelson:
@leonstr I have created https://github.com/mdjnelson/moodle-mod_customcert/issues/564.

PR #546, opened 2023-04-03
518 Trigger created, updated and deleted events

leonstr:
Add template_created, template_updated and template_deleted event classes, and trigger these as appropriate.

This is for issue #518. I think the main non-obvious changes are:

1. edit.php line 204: Add if ($tid) {} so $template->save() isn't called after template::create() as I think it's is redundant in this scenario, and it prevents both created and updated events being triggered.
2. template->copy_to_template() now takes the target instance as a parameter instead of just the ID. This is so that an event can be triggered for that target template.
3. copy_to_template() uses context to determine if it's a site-wide template being copied (needing a created event), or a template being loaded at course level (needing an updated event). This feels a bit messy.


mdjnelson:
Can you please rebase this and get all the tests passing, thanks. :)

mdjnelson:
I am working on this now and adding some extra events and unit tests.

mdjnelson:
Thanks. I ended up writing a fair few more events and wrote unit tests to ensure we aren't breaking anything. This took some time unfortunately which if I had known would not have started the work as there is a lot in the backlog. Oh well, it is done now. :)

PR #548, opened 2023-04-17
Fix Unit test

x-iy:
Fix  for failing PHPUnit test on 4.1

Undefined property: stdClass::$lang


mdjnelson:
Thanks, was able to replicate the error in the unit tests.

Issue #550, opened 2023-04-24
Maybe there should only be one background image


Julian-Tovar:
Good day team.

I've been working on `mod_certificate`, and I was asked to code it in such a way that only one background image can be added.

So that my code doesn't fully go to waste, and maybe it can be used here, I wanted to ask, does this make sense? I think it does, because when you add a second or third background images, they stack on top of each other and hide the backgrounds below. Besides, if the user wants to compose images, they can always add an "Image" element to the certificate, to put images on top of the background image (those images can then be moved and their size can be changed).

mdjnelson:
Ignore my previous comment. It's a sane idea just the PR needs adjusting. I left a comment there.

mdjnelson:
Closing. We may have a certificate with multiple pages and different backgrounds for each.

mdjnelson:
To make this sane we would have to add a check in `customcertelement_bgimage::can_add()` to check what page it was being added to and to also check if a bgimage existed. IMO it's too much effort.

PR #553, opened 2023-04-28
Removed background image from dropdown when there is one already (#550)

Julian-Tovar:
This PR creates a new variable `$skipbackground` to skip adding the background image to the element dropdown, when there is already one background image in a template

### DIFF
--- a/classes/edit_form.php
+++ b/classes/edit_form.php
@@ -241,6 +241,7 @@ class edit_form extends \moodleform {
         $mform->addHelpButton('pagerightmargin_' . $page->id, 'rightmargin', 'customcert');
 
         // Check if there are elements to add.
+        $hasbackground = false;
         if ($elements = $DB->get_records('customcert_elements', array('pageid' => $page->id), 'sequence ASC')) {
             // Get the total number of elements.
             $numelements = count($elements);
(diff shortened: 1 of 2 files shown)


classes/edit_form.php
```diff
@@ -257,6 +258,9 @@ protected function add_customcert_page_elements($page) {
                 $row = new \html_table_row();
                 $row->cells[] = $OUTPUT->render($elementname);
                 $row->cells[] = $element->element;
+                if (!$hasbackground && $element->element === 'bgimage') {
```
Julian-Tovar, line 261: This should only affect the 'bgimage' element.

mdjnelson:
Thanks for proposing a solution but I do not think it's a great idea to hardcode 'bgimage' in the core code. A proper solution would require a new method in `element/image/classes/element.php` that elements can override to return false if multiple instances are not allowed. You would then call this function in the core code and if it's false check before allowing it to be added.

mdjnelson:
Sorry, will reopen to allow for a new patch to be submitted.

mdjnelson:
I found a method I had forgotten about called `public statuc function can_add()` in the base class - you could probably override this in the background element.

Julian-Tovar:
Hi! Thank you so much @mdjnelson I really like the idea of using a base method that elements can override to return false if multiple instances are not allowed

mdjnelson:
Closing, see https://github.com/mdjnelson/moodle-mod_customcert/issues/550 for reason why.

Issue #558, opened 2023-05-30
Exception - TCPDF ERROR when mod/customcert/edit.php 'Save changes and preview' AND debugging NOTICES on


x-iy:
With debugging on, receive following error when trying to preview edited template:

```
Exception - TCPDF ERROR: Some data has already been output, can't send PDF file
More information about this error
    Debug info:  
Error code: generalexceptionmessage    
    Stack trace:  
line 3007 of /lib/tcpdf/tcpdf.php: Exception thrown
line 7698 of /lib/tcpdf/tcpdf.php: call to TCPDF->Error()
line 190 of /lib/pdflib.php: call to TCPDF->Output()
line 348 of /mod/customcert/classes/template.php: call to pdf->Output()
line 232 of /mod/customcert/edit.php: call to mod_customcert\template->generate_pdf()        

Output buffer:  <br /> <b>Notice</b>:  Trying to get property 'language' of non-object in <b>/var/www/xxxxxx/mod/customcert/classes/template.php</b> on line <b>275</b><br /> <br /> <b>Notice</b>:  Trying to get property 'language' of non-object in <b>/var/www/xxxxxx/mod/customcert/classes/template.php</b> on line <b>335</b><br />      
```
In this situation, $customcert = FALSE as `id` doesn't exist therefore `$customcert->language` is property of non-object and Notice created for display.

Suggested fix: check existence of `$customcert->language`  beforehand in `classes/template.php` lines 275 and 335

With debugging off, no exception occurs.

mdjnelson:
Thanks! This was fixed in https://github.com/mdjnelson/moodle-mod_customcert/commit/a851ff6c048b95ebbba051485aa2b482b9977e97. :)

x-iy:
Hi Mark,
Sorry for not being clear, but this is a separate issue to the one you mention as fixed.

This issue only occurs when:
- adding / editing a template via mod/customcert/manage_templates.php
- debugging set to: DEVELOPER: extra Moodle debug messages for developers
- clicking 'Save changes and preview' button

The issue doesn't occur when editing from a course activity.

I will create a MR with my suggested fix.

mdjnelson:
Thanks. I was able to replicate this, thanks for the PR. Ill comment over there.

PR #559, opened 2023-05-31
#558 add check for $customcert in classes/template.php

x-iy:
#558 proposed fix


mdjnelson:
I am going ahead and cherry-picking these then will push, thanks.

mdjnelson:
Done, thanks!

PR #572, opened 2023-07-16
571 Delete customcert_pages rows after elements

leonstr:
Previously template->delete_page() deleted the corresponding mdl_customcert_pages row, and then called element->delete() for each element on the page.  But the latter then called
element_deleted::create_from_element() which relied on that mdl_customcert_pages row to obtain the contextid for the event.

Now element->delete() is called for each element on the page, then the mdl_customcert_pages row is deleted.


mdjnelson:
Cherry-picked, thanks!

Issue #565, opened 2023-06-09
Mobile app no longer works


mdjnelson:
See https://moodle.org/mod/forum/discuss.php?d=447275#p1798669.

mdjnelson:
Pushed.

mdjnelson:
The reason for this issue is https://tracker.moodle.org/browse/MDL-76583, so fix is only on the `MOODLE_402_STABLE` branch.

PR #583, opened 2023-10-03
#565 Fix undefined external_format_text error in mobile app

dpalou:
This error was caused by this change in Moodle LMS:

https://tracker.moodle.org/browse/MDL-76583


mdjnelson:
Cherry-picked, thanks.

Issue #579, opened 2023-08-10
Dynamic field does not appear when attempting to reposition


carlicockrell:
On latest release of 4.0, dynamic fields do not appear so they are unable to be repositioned.

Video of issue: https://www.loom.com/share/cf117367cdc24c58a2e737fd2db5383c?sid=5a37bff0-f11c-4b4b-9062-c1da535c0eac

carlicockrell:
Hello,

I'm wondering if there is an update to this issue as it is currently resulting in a dealbreaker in using this tool as the dynamic fields are key feature. Thank you.

mdjnelson:
Fixed.

carlicockrell:
@mdjnelson I apologize for asking here, but we were wondering when the directly would be updated? https://moodle.org/plugins/mod_customcert Our AWS environment populates from the community and a new version has to be pushed for us to use this as we don't pull from git (causes update errors down the road). Is it possible to get 4.0.6 added to the community? Thank you.

mdjnelson:
Hi @carlicockrell there are a few more issues I want to deal with before doing a release. It's all planned work in the project board.

PR #589, opened 2023-12-15
579 Preview shortname if course custom field empty

leonstr:
Proposed fix for #579. Would need applying to all supported branches, for example, both MOODLE_402_STABLE and MOODLE_401_STABLE.


leonstr:
Replaced previous proposed change with one which shows the custom field's shortname instead of customfield_shortname. This is more consistent with the way "idnumber" is shown if course's ID number isn't populated.

Hope this is clear/didn't break anything/cause too much confusion.

mdjnelson:
Cherry-picked, thanks!

Issue #570, opened 2023-07-14
More issues with newly added events


leonstr:
I've tested the fixes for  #564  and see the following issues:

1. Unexpected _template_updated_ on create template (but not add activity instance). It's unexpected because it's in addition to the expected _template_created_ event. 
2. No event on clicking **Save changes** or **Add element** on **customcert/edit.php**. The latter action implicitly saves template changes, e.g. any name change. I'd expect a _template_updated_.
3. Event descriptions when adding, updating and deleting a course activity instance show "created the certificate template", "updated the certificate template" and "deleted the certificate template". Whereas these should say "created the certificate", "updated the certificate" and "deleted the certificate" respectively, with "... template" only displayed when adding, updating and deleting templates under **Site administration** > **Plugins** > **Activity modules** > **Custom certificate** > **Custom certificate settings**, **Manage templates**).
4. Loading a template for a course activity instance includes a _page_created_ event with context: system, whereas surely this should be context: cmid.

Tested with 7ee063e (MOODLE_401_STABLE) on Moodle 4.1.3+ (build: 20230526).

leonstr:
As a first go at fixing these I've [added branch 564-401 commit 1941274](https://github.com/mdjnelson/moodle-mod_customcert/compare/MOODLE_401_STABLE...leonstr:moodle-mod_customcert:564-401?expand=1).

Another issue: I don't like that every _page_updated_ also has a _template_updated_, if there's one action then two events seems wrong. Having one _template_updated_ + one or more _page_updated_ would make sense. I'd need to do a bit more testing to see if the _template_updated_ could be removed from **template->save_page()**, i.e. does a _template_created_ or _template_updated_ definitely get triggered every time **save_page()** is called? I haven't tried to fix this.

leonstr:
Another issue: There are two _template_created_ events when copying a template. Possibly **template->copy_to_template()** doesn't need the:

```
                \mod_customcert\event\template_created::create_from_template($copytotemplate)->trigger();
```

because **manage_templates.php** calls **template->create()** which calls:

```
        \mod_customcert\event\template_created::create_from_template($template)->trigger();
```

?

leonstr:
There's a further issue that rearranging elements by drag and drop then saving these this doesn't trigger any events. At this stage I think this specific issue should be added to it's own ticket, patching the above issues has taken long enough and retro-fitting events to `rearrange.php` will take a bit of thought.

mdjnelson:
[...]

PR #595, opened 2024-01-10
Ensure events are only triggered when necessary

mdjnelson:
1. Suppress template_updated when creating new template.

2. Fix Save changes/Add element so template_updated triggered if name changed.

3. Fix context for page_created when loading a template in course instance (was system now course module).

### DIFF
--- a/classes/template.php
+++ b/classes/template.php
@@ -75,7 +75,7 @@ class template {
         $DB->update_record('customcert_templates', $savedata);
 
         // Only trigger event if the name has changed.
-        if ($savedata->name != $data->name) {
+        if ($this->get_name() != $data->name) {
             \mod_customcert\event\template_updated::create_from_template($this)->trigger();
         }
     }
@@ -155,9 +155,9 @@ class template {
                     // Update the page.
                     $DB->update_record('customcert_pages', $p);
 
+                    // Calling code is expected to trigger template_updated
+                    // after this method.
                     \mod_customcert\event\page_updated::create_from_page($p, $this)->trigger();
-
-                    \mod_customcert\event\template_updated::create_from_template($this)->trigger();
                 }
             }
         }
@@ -171,27 +171,11 @@ class template {
     public function delete() {
         global $DB;
 
-        // Delete the elements.
-        $sql = "SELECT e.*
-                  FROM {customcert_elements} e
-            INNER JOIN {customcert_pages} p
-                    ON e.pageid = p.id
-                 WHERE p.templateid = :templateid";
-        if ($elements = $DB->get_records_sql($sql, array('templateid' => $this->id))) {
-            foreach ($elements as $element) {
-                // Get an instance of the element class.
-                if ($e = \mod_customcert\element_factory::get_element_instance($element)) {
-                    $e->delete();
-                } else {
-                    // The plugin files are missing, so just remove the entry from the DB.
-                    $DB->delete_records('customcert_elements', array('id' => $element->id));
-                }
-            }
-        }
-
         // Delete the pages.
-        if (!$DB->delete_records('customcert_pages', array('templateid' => $this->id))) {
-            return false;
+        if ($pages = $DB->get_records('customcert_pages', array('templateid' => $this->id))) {
+            foreach ($pages as $page) {
+                $this->delete_page($page->id, false);
+            }
         }
 
         // Now, finally delete the actual template.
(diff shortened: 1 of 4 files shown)


mdjnelson:
Hi @leonstr, sorry it took me a while to see you had pushed code but there was no PR yet, so created this. Does this include all fixes for events or is there still more to do? I would like to fix up the events code before I push another release.

leonstr:
@mdjnelson From memory this is not yet complete. Apologies for the delay. I'll aim to pick this up in the next week with a view to getting this wrapped up.

mdjnelson:
Thanks @leonstr!

mdjnelson:
Ping @leonstr! :)

leonstr:
Just acknowledging the ping. I've been working on this on Saturday and today. A bit more testing is required, I'll try to have this done in the next day.

leonstr:
Have updated PR. Hopefully this fixes all issues listed in #570 and doesn't break anything (PHPUnit tests pass on my host).

mdjnelson:
Looking now 👀.

classes/template.php
```diff
@@ -75,7 +75,7 @@ public function save($data) {
         $DB->update_record('customcert_templates', $savedata);
 
         // Only trigger event if the name has changed.
-        if ($savedata->name != $data->name) {
```
mdjnelson, line 78: So this was never being triggered since they would be the same. I think this needs a unit test to verify it is getting triggered then we would have caught this. Also, please use the method `get_name()`. *sigh* Looking at this plugin's code I wish I could go back in time and rewrite it but I guess thats software development for you!

load_template.php
```diff
@@ -51,22 +51,16 @@
 // Check that they have confirmed they wish to load the template.
 if ($confirm && confirm_sesskey()) {
     // First, remove all the existing elements and pages.
-    $sql = "SELECT e.*
-              FROM {customcert_elements} e
-        INNER JOIN {customcert_pages} p
-                ON e.pageid = p.id
+    $sql = "SELECT p.*
```
mdjnelson, line 54: This SQL seems to be just hanging around for no reason and is not used.

load_template.php
```diff
@@ -51,22 +51,16 @@
...
-            // Get an instance of the element class.
-            if ($e = \mod_customcert\element_factory::get_element_instance($element)) {
-                $e->delete();
-            }
-        }
-    }
 
     // Delete the pages.
-    $DB->delete_records('customcert_pages', array('templateid' => $template->get_id()));
+    if ($pages = $DB->get_records('customcert_pages', array('templateid' => $template->get_id()))) {
```
mdjnelson, line 59: Please use [] instead of array().

leonstr:
Thanks for the feedback @mdjnelson most of these look straightforward, give me another day or so to look at what PHPUnit testing I can add.

mdjnelson:
No worries, thanks for you help! I am currently doing a refactor to make pages, templates and elements persistents which can contain the logic for triggering these events (and not just that but this was needed a long time ago cause the logic is all over the place). However, this will take me a bit so go ahead with what you are doing and yeh, unit tests would be great to know my refactor isnt screwing everything up. :D

classes/template.php
```diff
@@ -157,7 +157,6 @@ public function save_page($data) {
 
                     \mod_customcert\event\page_updated::create_from_page($p, $this)->trigger();
 
-                    \mod_customcert\event\template_updated::create_from_template($this)->trigger();
```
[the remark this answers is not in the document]
leonstr: Added a comment to hopefully make this a bit clearer.

classes/template.php
```diff
@@ -208,8 +191,10 @@ public function delete() {
      * Handles deleting a page from the template.
      *
      * @param int $pageid the template page
+     * @param bool $triggertemplateupdatedevent False if page is being deleted
+     * during deletion of template.
      */
-    public function delete_page($pageid) {
+    public function delete_page($pageid, $triggertemplateupdatedevent = true) {
```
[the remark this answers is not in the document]
leonstr: Done.

classes/template.php
leonstr: Added PHPUnit tests test_updating_a_template and test_updating_a_template_no_change.

load_template.php
leonstr: Removed unused SQL.

load_template.php
leonstr: Done.

leonstr:
PHPUnit tests: I've added checks for events' contexts, currently this is "system" except for test_load_template which I've added to catch one of the problems originally reported (Loading a template for a course activity instance includes a page_created event with context: system). I've also removed the objectid checks in test_copy_to_template since a) these were incorrect (param1 and param2 are the same), and b) we're using `copy_to_template()` to create records so don't deterministically know what their IDs will be (likely the existing IDs + 1, but we shouldn't count on that).

mdjnelson:
Cherry-picked, thanks!

Issue #599, opened 2024-01-25
No element_updated/template_updated event after reposition elements


leonstr:
Using "Reposition elements" (`mod/customcert/rearrange.php`) and clicking _Save and close_ or _Save and continue_ does not trigger any element_updated or template_updated events.

mod_customcert 2022112804 on Moodle 4.1.5+ (Build: 20230929).

leonstr:
Added PR #600 for this.

mdjnelson:
Push to git, thanks!

PR #600, opened 2024-01-25
Add updated events for reposition elements (#599)

leonstr:
When using drag and drop to reposition elements trigger element_updated and template_updated events on save.

### DIFF
--- a/classes/event/element_updated.php
+++ b/classes/event/element_updated.php
@@ -88,6 +88,23 @@ class element_updated extends \core\event\base {
     }
 
     /**
+     * Create instance of event for the specified element.
+     *
+     * @param int $elementid ID of the element.
+     * @param \mod_customcert\template $template Template containing the above
+     * element.
+     * @return element_updated
+     */
+    public static function create_from_id(int $elementid, \mod_customcert\template $template): element_updated {
+        $data = [
+            'contextid' => $template->get_contextid(),
+            'objectid' => $elementid,
+        ];
+
+        return self::create($data);
+    }
+
+    /**
      * Returns relevant URL.
      * @return \moodle_url
      */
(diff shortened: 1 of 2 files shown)


classes/event/element_updated.php
```diff
@@ -87,6 +87,26 @@ public static function create_from_element(\mod_customcert\element $element) : e
...
     }
 
+    /**
+     * Create instance of event for the specified element.
+     *
+     * @param int $elementid ID of the element.
+     * @param int $templateid ID of template containing the above element.
+     * @return element_updated
+     */
+    public static function create_from_ids(int $elementid, int $templateid) : element_updated {
```
mdjnelson, line 97: To conform to PSR-12 standards can you please set this to `int $templateid): element_updated` (notice no spacing after the method declaration). Thanks!

leonstr:
Patch revised remove extraneous space in method declaration to conform to PSR-12 standards.

classes/event/element_updated.php
```diff
@@ -87,6 +87,26 @@ public static function create_from_element(\mod_customcert\element $element) : e
...
+     * Create instance of event for the specified element.
+     *
+     * @param int $elementid ID of the element.
+     * @param int $templateid ID of template containing the above element.
+     * @return element_updated
+     */
+    public static function create_from_ids(int $elementid, int $templateid): element_updated {
+        global $DB;
+
+        $template = $DB->get_record('customcert_templates', ['id' => $templateid]);
```
mdjnelson, line 100: To avoid having do this query why not pass in the template object that is created above? It will save you a query.

leonstr: Thanks for the feedback, revised patch with the suggested improvement.

mdjnelson: Thanks.

mdjnelson:
Push to git, thanks!

Issue #531, opened 2023-02-02
CPU % runs high even on an idle server caused by Custom Cert email task


DeRespinis:
Moodle 4.1 and Moodle 4.0
Custom Cert Version 4.02  2022041902

CPU % runs high even on an idle server caused by Custom Cert email task.

The CPU % ranges from 0% to 3% on an idle machine with the Custom Cert email task disabled.
Mostly sits at 0%

The CPU % ranges from 8% - 15% on an idle machine with the Custom Cert email task enabled.
The CPU never stops and operates at a minimum of 8%

I noticed the change when I updated from 3.7 to 4.0.    I went from 3.7 to 3.9 to 3.11 then to 4.0 and 4.1.

The Email task does not report any errors but the time to complete each task is taking about 20 seconds.
 
![image](https://user-images.githubusercontent.com/6350928/216445289-b9ffc06d-9e80-471e-803b-68bb8aaad172.png)

I have this running on both an 4-5 year old AWS Linux 2 LAMP and a new AWS Bitnami Moodle site. 

This does not seem like correct behavior.

Thank for your help.

mdjnelson:
It's a pretty complex task that tries to achieve a lot. I would be happy to merge any code that improves this task (maybe adding some caching?).

Peterburnett:
Hi Folks,

I've stumbled here after an upgrade to 4.1 from 3.9, where I am seeing the email task use 4 million DB queries every run, while not actually issuing any certs at all.

![image](https://github.com/mdjnelson/moodle-mod_customcert/assets/21162968/9f11c9cf-1eb5-4092-85ef-8785bbd47913)

After digging around on the Moodle tracker, it appears that 3.10 changed get_dynamic_info and introduced a fairly heavy overhead.

https://tracker.moodle.org/browse/MDL-72402

After doing a perf dump using tool_excimer, the flamegraph is basically all just stuck calculating uservisible here:

https://github.com/mdjnelson/moodle-mod_customcert/blob/MOODLE_402_STABLE/classes/task/email_certificate_task.php#L127

![Pasted image 20240117141558](https://github.com/mdjnelson/moodle-mod_customcert/assets/21162968/c472c5f7-c222-44df-8556-f53838ea971e)

[...]

PR #610, opened 2024-03-13
Improvement to the speed of the emailing certificate task (#531)

mdjnelson:


mdjnelson:
@mohamedmohamedatia can we not use the `task_log` table to determine when it was last run instead of introducing another table? It would be a lot nicer and easier to maintain than us implementing our own logic. Then you just need to check the issued times and only get the issues that were done after the last time the task was run.

Small notes - 

1. Do not use `array()`, please use the short syntax `[]`.
2. Align the settings file settings at the left of the page. Make it consistent with the rest of the file.
3. There is no need for language strings to have capitals in all the words. The string `Include Certificates in Not Visible Courses` should be `Include certificates in not visible courses`. Also in this case `not visible` can be changed to `hidden`.
4. `$fastmoduleinfo = get_fast_modinfo($customcert->courseid, $enroluser->id)->instances['customcert'][$customcert->id];` - `$enroluser` is not declared.

Thanks, look forward to your reply.

mohamedmohamedatia:
@mdjnelson

Apologies for the delay in responding. I missed reading the notification email and just noticed it now. Thank you for your valuable feedback.

I'll review and implement your suggestions promptly and will reply soon.

Best regards,

mohamedmohamedatia:
@mdjnelson
I appreciate your input and I'd like to clarify a few points about the changes I've implemented.

**Before my edits,** the task was indeed handling two primary functions: **issuing certificates** and **sending emails**. It loops through **all certificates** on the system to check if new users could issue them.

I introduced a new feature that allows administrators to determine how many certificates to process during each run. By default, **if set to 0, it will check all certificates.**

I introduced a table to store **the starting position for certificate processing** each run. For example, if there are 1000 certificates and the admin sets 100 certificates per run, after the first run, it stores 100 as the last_processed field. Subsequent runs begin from position 101, _**I am not sure if the task_log table alone can achieve this, I will wait for further direction from your side if there is a better way to store the start position.**_

Regarding Other notes – I created a branch named Notes#610 and **all notes are done**; should I merge now?

Regarding your concern about the count, it's for **looping over certificates, not users**. It determines the batch size for processing certificates, not users for issuance, as highlighted above.

Regarding users, instead of iterating over **all enrolled users**, the loop now targets **users with the capability "issue certificate"**.
For instance, if a certificate is restricted to a certain role, such as "teacher," only users with that role will be included in the loop. Similarly, if the certificate is tied to completing a particular activity, only users who have completed that activity will be considered.

Additionally, I added the ability for admins to **exclude certificates from hidden courses and categories**, vital in our university where old terms are often hidden. This ensures **active term certificates are only processed** by default, with the option to include hidden courses if needed. Also, admins can **exclude old courses that ended in the past over a determined period from processing.**

mdjnelson:
@mohamedmohamedatia you can just add a commit onto the same branch instead of creating a new pull request or branch and I can see them here. I'll address the rest of this review when I am next working on the project.

mdjnelson:
Thanks @mohamedmohamedatia. Can you squash these commits into one and rebase this on the latest version of MOODLE_402_STABLE. Right now there are conflicts. Thanks!

mohamedmohamedatia (approved):

mohamedmohamedatia:
Hi Mark,
Thank you for your guidance. I've squashed the commits as requested and rebased the branch onto the latest version of MOODLE_402_STABLE

It was my first time squashing commits, but I followed online tutorials to ensure I did it correctly. As a result, we now have two succinct commits:

1. Configurable settings for task efficiency
2. Optimize email certificate task by reducing database reads/writes

Please let me know if there's anything else I can assist you with.
Thanks again for your direction!

mdjnelson:
Hi @mohamedmohamedatia, did you pull down the latest updates of the customcert version 4.2 and then do the rebase as I still get the errorr "This branch has conflicts that must be resolved" on GitHub? Also these aren't exactly succinct since the second commit contains fixes for the first commit. I would prefer them separated, or just have one commit (easiest solution).

mdjnelson:
@mohamedmohamedatia, there are still conflicts. You will need to `git pull` the `MOODLE_402_STABLE`, then checkout your branch and do a rebase on it. See https://github.com/mdjnelson/moodle-mod_customcert/commits/MOODLE_402_STABLE/ - this is the latest.

mohamedmohamedatia:
Hi  Mark,
I've successfully addressed the conflicts and combined the changes into a single commit. is everything OK now?

mdjnelson:
Awesome work @mohamedmohamedatia. When I get time ill give it another review. :)

mdjnelson:
Hi @mohamedmohamedatia can you fix this issues raised above by the automated tests? :)

mohamedmohamedatia:
Hi Mark, Noted, will address them soon.

mohamedmohamedatia:
Hi Mark,
I fixed the test issues. All tests now pass. Squashed commits into one.
Results: OK (7 tests, 29 assertions).

mdjnelson:
Thanks @mohamedmohamedatia,

I am currently busy with work outside of Moodle but when I get a chance ill give it an extensive review.

Regards,

Mark

mdjnelson:
Looking at this now. Just going to comment that this will solve #531 so it links in GitHub.

db/install.php
```diff
@@ -0,0 +1,52 @@
...
+    $dbman = $DB->get_manager();
+
+    // Add a default row to the customcert_task_progress table.
+    $defaultdata = new stdClass();
+    $defaultdata->taskname = 'email_certificate_task';
+    $defaultdata->last_processed = 0;
+    $defaultdata->total_certificate_to_process = 0;
+
+    // Write close to ensure the transaction is committed.
+    \core\session\manager::write_close();
```
mdjnelson, line 47: Why is this needed?

db/install.php
```diff
@@ -0,0 +1,52 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Customcert module upgrade code.
+ *
+ * @package    mod_customcert
+ * @copyright  2016 Mark Nelson <markn@moodle.com>
```
mdjnelson, line 21: This should be `2024 Mohamed Atia <matia12@gmail.com>`.

db/install.php
```diff
@@ -0,0 +1,52 @@
+<?php
```
mdjnelson, line 1: Is this file *really* needed? I am reluctant to add a new file. Could you simply not check if there were no rows then assume the variables are `0`? I would prefer not adding more code than is needed, especially a whole new file that manipulates the database.

classes/task/email_certificate_task.php
```diff
@@ -248,8 +311,18 @@ public function execute() {
                 }
 
                 // Set the field so that it is emailed.
-                $DB->set_field('customcert_issues', 'emailed', 1, ['id' => $user->issueid]);
+                $issueids[] = $user->issueid;
+            }
+            if (!empty($issueids)) {
+                $DB->set_field_select('customcert_issues', 'emailed', 1, 'id IN (' . implode(',', $issueids) . ')');
```
mdjnelson, line 317: You need to use `$DB->get_in_or_equal` here. See example at https://github.com/mdjnelson/moodle-mod_customcert/blob/MOODLE_402_STABLE/classes/certificate.php#L467.

classes/task/email_certificate_task.php
```diff
@@ -111,27 +165,35 @@ public function execute() {
...
+            // Get the context of the Custom Certificate module.
+            $cm = get_coursemodule_from_instance('customcert', $customcert->id, $customcert->course);
+            $context = \context_module::instance($cm->id);
 
-                // Don't want to email those with the capability to manage the certificate.
-                if (has_capability('mod/customcert:manage', $context, $enroluser->id)) {
+            // Now, get a list of users who can view and issue the certificate but have not yet.
+            // Get users with the mod/customcert:receiveissue capability in the Custom Certificate module context.
+            $userswithissue = get_users_by_capability($context, 'mod/customcert:receiveissue',
+            'u.id, username, firstname, lastname, email, firstnamephonetic, lastnamephonetic, middlename, alternatename');
```
mdjnelson, line 178: This is prone to have errors in the future if these fields ever change and we are hardcoding them. Please use `\core_user\fields::for_identity($context)->get_sql('u', true, '', '', false);`. You might have to look into this more as I am not too familiar with this API - see https://docs.moodle.org/dev/User_fields.

classes/task/email_certificate_task.php
```diff
@@ -111,27 +165,35 @@ public function execute() {
...
+            // Get users with mod/customcert:view capability.
+            $userswithview = get_users_by_capability($context, 'mod/customcert:view',
+            'u.id, username, firstname, lastname, email, firstnamephonetic, lastnamephonetic, middlename, alternatename');
```
mdjnelson, line 181: Same as above.

classes/task/email_certificate_task.php
mdjnelson: In fact - does this even need to be included? Why not just leave the default?

classes/task/email_certificate_task.php
```diff
@@ -49,27 +49,81 @@ public function get_name() {
...
 
         // The renderers used for sending emails.
         $page = new \moodle_page();
         $htmlrenderer = $page->get_renderer('mod_customcert', 'email', 'htmlemail');
         $textrenderer = $page->get_renderer('mod_customcert', 'email', 'textemail');
-        foreach ($customcerts as $customcert) {
+
+        // Store the total count of certificates in the database.
+        $totalcertificatestoprocess = count($customcerts);
+        $DB->set_field('customcert_task_progress', 'total_certificate_to_process', $totalcertificatestoprocess, [
```
mdjnelson, line 104: The field is set here but it doesn't actually do anything. It just gets set then forgotten about afaict.

lang/en/customcert.php
```diff
@@ -229,3 +229,12 @@
 
 // Acess API.
 $string['customcert:managelanguages'] = 'Manage language on edit form';
+
+$string['certificatesperrun'] = 'Certificates per run';
+$string['certificatesperrun_desc'] = 'Enter the number of certificates to process per scheduled task run <b>where 0 means it will process all certificates</b>.';
+$string['includeinnotvisiblecourses'] = 'Include certificates in hidden courses';
+$string['includeinnotvisiblecourses_desc'] = 'Check this box to include certificates in courses that are not visible to the user.';
+$string['certificateexecutionperiod'] = 'Certificate execution period';
+$string['certificateexecutionperiod_desc'] = 'Specify the period for which certificates should be executed based on their end date. <b>Set to 0 to execute all certificates, regardless of their age.</b>';
```
mdjnelson, line 238: Please remove the bold tags.

lang/en/customcert.php
```diff
@@ -229,3 +229,12 @@
 
 // Acess API.
 $string['customcert:managelanguages'] = 'Manage language on edit form';
+
+$string['certificatesperrun'] = 'Certificates per run';
+$string['certificatesperrun_desc'] = 'Enter the number of certificates to process per scheduled task run <b>where 0 means it will process all certificates</b>.';
```
mdjnelson, line 234: Please remove the bold tags.

settings.php
```diff
@@ -58,14 +58,34 @@
...
+    get_string('scheduledtaskconfigdesc', 'customcert')));
+
+$settings->add(new admin_setting_configtext('customcert/certificatesperrun',
+    get_string('certificatesperrun', 'customcert'),
+    get_string('certificatesperrun_desc', 'customcert'),
+    0, PARAM_INT));
+$settings->add(new admin_setting_configcheckbox('customcert/includeinnotvisiblecourses',
+    get_string('includeinnotvisiblecourses', 'customcert'),
+    get_string('includeinnotvisiblecourses_desc', 'customcert'), 0));
+$settings->add(
```
mdjnelson, line 73: I am not convinced this setting is needed. It's confusing. What is a certificate end date? It's just randomly the course end date which isn't clear. Also, a course may end but still want to send out certificates, eg. it ends after the final exam, then its closed, teacher does grading and then the user gets the certificate - valid scenario. I say we should remove this setting completely. Thoughts?

settings.php
```diff
@@ -58,14 +58,34 @@
...
+
+$settings->add(new admin_setting_heading('scheduledtaskconfig',
+    get_string('scheduledtaskconfigheading', 'customcert'),
+    get_string('scheduledtaskconfigdesc', 'customcert')));
+
+$settings->add(new admin_setting_configtext('customcert/certificatesperrun',
+    get_string('certificatesperrun', 'customcert'),
+    get_string('certificatesperrun_desc', 'customcert'),
+    0, PARAM_INT));
+$settings->add(new admin_setting_configcheckbox('customcert/includeinnotvisiblecourses',
```
mdjnelson, line 70: Again, I am not convinced this is needed. If the course is not visible to the user then they won't be able to view it anyways so why define it here?

ericmerrill:
I don't want to step on any toes, but I (or someone on my team) could review this, and update code with Mark's suggestions if @mohamedmohamedatia won't have time soon to work on this. We have several clients that have been having performance problems with this.

ojnadjarm:
Hi @mdjnelson, so I went deep into this issue and I think we have 2 pain points here. 
1) The main query, brings ALL the certificates that have the emailing active, no matter how old the certificate is or if still active. This should not be an issue for 100 certificates but with site with 1000+ or event 10000+ this doesn't scale. We are maybe loopping on thousands of certificates that probably don't need to be processed. So Excluding hidden courses and old courses directly on the query it will help a lot because we do a lot of queries per certificate, I think even more than running the query by a batches. We should still mantain that posibility because if there are a lot of active certificates It will be needed, also and I added a filter not only for inactive courses but also for courses that doesn't have an enddate but the last issued certificate is older than what is configured by the user.(Let say the lastest issued certificate is 2 years old, we probably should ignore that certificate).
2) Issuing the certificates and emailing on the schedule task could be a really heavy load when we are doing a bulk issuing. I separate the issues from the emailing, It was not only creating the issues, but creating the PDF and sending the email, too much work for a single task. I added a new shechedule task that will only issue the certificate and if an email is requested it will created an adhoc task that will be the one on creating and emailing the certificate. This will be enable by a setting. If not enable it will do it live. This is the PR I create to our repo. Let me know what do you think if you are ok with it we can create a PR to the main repo.
https://github.com/moodleus/moodle-mod_customcert/pull/1/files

mdjnelson:
Let me look at this properly tomorrow. I appreciate you taking your time to work on this and contribute. No harm in doing a PR now so we can start a discussion there.

mdjnelson:
I also noticed from this fork https://github.com/Claud10R/moodle-mod_customcert that @Claud10R created some commits to work on this. Would be good to see if perhaps we can make this even more efficient. Waiting on PR before reviewing. :)

ojnadjarm:
Hi, awesome I'll review it CLaud's commits.

settings.php
mohamedmohamedatia: Hi Mark,

I apologize for the delay in responding as I was on an extended leave for nearly two months and did not review emails during that time.

Regarding the hidden categories and courses, I configured them to allow administrators to tailor settings according to their organization's needs. Different organizations have varied operational methods. For instance, some universities have between 200K to 500K enrollments per term and retain courses on Moodle for 3 to 10 years. These older courses are usually hidden and are rarely accessed for certificate issuance. Therefore, it might not be efficient to include these in certificate checks for the entire 10 years. Allowing administrators to manage this through configurations ensures flexibility and efficiency.

Similarly, for old courses, I provided options for administrators to decide the time frame for system checks. For example, the system could be set to check only the courses from the last six months. If there is a need to process older courses, administrators can adjust the configurations as necessary.

I hope this addresses your concerns regarding the settings. If there are any other issues you are referring to, please let me know.

Best regards,

settings.php
mohamedmohamedatia: Before my edits, the plugin would check and process hidden courses. I changed this to a configurable setting because, in most cases, certificates are not issued for users in old, hidden courses. However, if there's a need to process everything, such as when a support ticket requires reissuing all certificates from previous terms, administrators still have the capability to do so. Does that make sense?

classes/task/email_certificate_task.php
mohamedmohamedatia: This Field is used when Custom certificate settings > Scheduled task configuration > Certificates per run is set to any value other than Zero.

classes/task/email_certificate_task.php
mohamedmohamedatia: Agree with you, I removed them

classes/task/email_certificate_task.php
mohamedmohamedatia: removed

classes/task/email_certificate_task.php
```diff
@@ -248,8 +311,18 @@ public function execute() {
                 }
 
                 // Set the field so that it is emailed.
-                $DB->set_field('customcert_issues', 'emailed', 1, ['id' => $user->issueid]);
+                $issueids[] = $user->issueid;
+            }
+            if (!empty($issueids)) {
```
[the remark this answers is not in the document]
mohamedmohamedatia: done

classes/task/email_certificate_task.php
mohamedmohamedatia: Done

db/install.php
mohamedmohamedatia: While I agree that minimizing code is ideal, directly setting default values within the table structure during installation is not possible in Moodle. 

I've implemented default values in the upgrade process.

However, for fresh installations, the table will be created without initial data. Are there alternative approaches to consider?

db/install.php
mohamedmohamedatia: Done

db/install.php
```diff
@@ -0,0 +1,52 @@
...
+ * @return bool always true
+ */
+
+/**
+ * Custom code to be run on installing the plugin.
+ */
+function xmldb_customcert_install() {
+    global $DB;
+
+    $dbman = $DB->get_manager();
```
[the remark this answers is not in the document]
mohamedmohamedatia: removed

db/install.php
mohamedmohamedatia: Removed

db/upgrade.php
```diff
@@ -235,5 +235,38 @@ function xmldb_customcert_upgrade($oldversion) {
         upgrade_mod_savepoint(true, 2023042405, 'customcert');
     }
 
+    if ($oldversion < 2023042408) {
```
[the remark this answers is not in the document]
mohamedmohamedatia: Done

db/upgrade.php
```diff
@@ -235,5 +235,38 @@ function xmldb_customcert_upgrade($oldversion) {
...
+
+        // Define table customcert_task_progress to be created.
+        $table = new xmldb_table('customcert_task_progress');
```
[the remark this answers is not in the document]
mohamedmohamedatia: Done, Yes I guess we should have task for email and task for issue certificate

db/upgrade.php
```diff
@@ -235,5 +235,38 @@ function xmldb_customcert_upgrade($oldversion) {
...
+
+        // Adding fields to table customcert_task_progress.
+        $table->add_field('id', XMLDB_TYPE_INTEGER, '10', null, XMLDB_NOTNULL, XMLDB_SEQUENCE, null);
+        $table->add_field('taskname', XMLDB_TYPE_CHAR, '255', null, XMLDB_NOTNULL, null, 'email_certificate_task');
+        $table->add_field('last_processed', XMLDB_TYPE_INTEGER, '20', null, XMLDB_NOTNULL, null, '0');
+        $table->add_field('total_certificate_to_process', XMLDB_TYPE_INTEGER, '20', null, XMLDB_NOTNULL, null, '0');
+
+        // Adding keys to table customcert_task_progress.
+        $table->add_key('primary', XMLDB_KEY_PRIMARY, ['id']);
+        $table->add_key('usermodified', XMLDB_KEY_FOREIGN, ['usermodified'], 'user', ['id']);
```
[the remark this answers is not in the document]
mohamedmohamedatia: I removed these forigen keys as they are not needed

db/upgrade.php
```diff
@@ -235,5 +235,38 @@ function xmldb_customcert_upgrade($oldversion) {
...
+        if (!$dbman->table_exists($table)) {
+            $dbman->create_table($table);
+            // Add a default row to the customcert_task_progress table.
+            $defaultdata = new stdClass();
+            $defaultdata->taskname = 'email_certificate_task';
+            $defaultdata->last_processed = 0;
+            $defaultdata->total_certificate_to_process = 0;
+
+            // Write close to ensure the transaction is committed.
+            \core\session\manager::write_close();
```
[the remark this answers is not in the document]
mohamedmohamedatia: Removed

db/upgrade.php
```diff
@@ -235,5 +235,38 @@ function xmldb_customcert_upgrade($oldversion) {
...
+
+            // Insert the default data into the table.
+            $DB->insert_record('customcert_task_progress', $defaultdata);
+        }
+        // Customcert savepoint reached.
+        upgrade_mod_savepoint(true, 2023042408, 'customcert');
```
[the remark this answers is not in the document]
mohamedmohamedatia: Done

db/upgrade.php
```diff
@@ -235,5 +235,38 @@ function xmldb_customcert_upgrade($oldversion) {
...
+        $table->add_field('total_certificate_to_process', XMLDB_TYPE_INTEGER, '20', null, XMLDB_NOTNULL, null, '0');
+
+        // Adding keys to table customcert_task_progress.
+        $table->add_key('primary', XMLDB_KEY_PRIMARY, ['id']);
+        $table->add_key('usermodified', XMLDB_KEY_FOREIGN, ['usermodified'], 'user', ['id']);
+
+        // Conditionally launch create table for customcert_task_progress.
+        if (!$dbman->table_exists($table)) {
+            $dbman->create_table($table);
+            // Add a default row to the customcert_task_progress table.
```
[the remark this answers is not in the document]
mohamedmohamedatia: Done

db/upgrade.php
```diff
@@ -235,5 +235,38 @@ function xmldb_customcert_upgrade($oldversion) {
...
+            $defaultdata->last_processed = 0;
+            $defaultdata->total_certificate_to_process = 0;
+
+            // Write close to ensure the transaction is committed.
+            \core\session\manager::write_close();
+
+            // Insert the default data into the table.
+            $DB->insert_record('customcert_task_progress', $defaultdata);
+        }
+        // Customcert savepoint reached.
```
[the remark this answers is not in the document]
mohamedmohamedatia: Done

db/upgrade.php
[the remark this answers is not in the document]
mohamedmohamedatia: I used customcert_email_task_prgrs due to 28 chars limit

lang/en/customcert.php
mohamedmohamedatia: Done

lang/en/customcert.php
mohamedmohamedatia: Done

settings.php
```diff
@@ -58,14 +58,34 @@
         new moodle_url('/mod/customcert/download_all_certificates.php'), ''));
 }
 
+
```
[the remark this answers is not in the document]
mohamedmohamedatia: Removed

mohamedmohamedatia:
Hi Mark,

I apologize for the delay in my response. I was on an extended leave for two months and was unable to review emails during that period.

I have reviewed your comments and made the necessary changes. I’ve also added my responses to each of your comments. Please note that while I haven’t completely halted the processing of hidden courses, I’ve enabled administrators to make a selection based on their operational policies. The default setting will not process hidden courses.

The default configuration now processes courses with an end date no older than one year. However, administrators can adjust this setting to zero if they wish to include all courses. Courses without an end date are considered open and will always be processed, such as self-study courses.

I will review #632 at a later time, but I suggest we proceed with #610 for now and consider #632 as future enhancements. Please let me know if you have any suggestions for alternative scenarios.

Best regards,

settings.php
mdjnelson: Yep, making it a setting is the way to go! Thanks.

settings.php
mdjnelson: Thanks @mohamedmohamedatia. Makes sense.

settings.php
```diff
@@ -58,14 +58,34 @@
...
+    );
+
 $settings->add(new admin_setting_heading('defaults',
     get_string('modeditdefaults', 'admin'), get_string('condifmodeditdefaults', 'admin')));
 
 $yesnooptions = [
     0 => get_string('no'),
     1 => get_string('yes'),
 ];
-
```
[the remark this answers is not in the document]
mdjnelson: Minor point, ignoring it.

mdjnelson:
Ignore the failures for now as it seems to be because of https://github.com/moodlehq/moodle-plugin-ci/issues/309.

mohamedmohamedatia:
No worries, I’ll ignore the failures for now since it seems to be related to moodlehq/moodle-plugin-ci#309. I ran the tests on my side, and everything passed just fine!

mdjnelson:
I have cherry-picked this (with some minor changes) in to 4.4, 4.3, 4.2 and 4.1. I have not released it as a version in the plugins DB though.

@mohamedmohamedatia just to confirm that if this processes a particular course that it does not ignore it when it does the batching? New students could still qualify for the certificate after its been processed.

mohamedmohamedatia:
Hi Mark
Great news, Thank you for cherry-picking the changes into the different
versions!, I can confirm that after a course is processed, it is not
excluded in future batches. This ensures that any new students who meet the
criteria after the initial processing will still qualify for the
certificate.

mdjnelson:
Awesome, closing. :) Thanks @mohamedmohamedatia for all your hard work, it's appreciated.

mohamedmohamedatia:
Thanks! It’s been great working on this plugin. I appreciate the chance to contribute!

mdjnelson:
Hi @mohamedmohamedatia, it seems this introduced an urgent bug that needs fixing. Can you take a look at https://moodle.org/mod/forum/discuss.php?d=461260 and see if you can spot the issue?

mdjnelson:
The following code was removed -

```
// Now check if the certificate is not visible to the current user.
$cm = get_fast_modinfo($customcert->courseid, $enroluser->id)->instances['customcert'][$customcert->id];
if (!$cm->uservisible) {
    continue;
}
```

mdjnelson:
I created a fix you can find at https://github.com/mdjnelson/moodle-mod_customcert/commit/94cfd4b130392e1a204844a0074c8938d09bf642. I am not sure why `$infomodule->filter_user_list($userswithissueview)` did not work, strange. I have created a unit test now so we wont run into this issue again (hopefully).

Issue #620, opened 2024-05-27
Non-editing teachers can see the download button in report but can't download anything (presents a blank page)


rhell4:
Viewing the report is controlled by the `mod/customcert:viewreport` permission but downloading a student's certificate is controlled by the `mod/customcert:manage` permission. There should be some consistency here and whatever is done for individual downloads should match the download all functionality.

PR #621, opened 2024-05-27
Issue #620: use viewreport cap for downloading student certs

rhell4:
As stated in issue #620 there are some inconsistencies in when the download button is shown in the report and when the user can actually download that cert, so non-editing teachers can see the buttons but get sent to a blank page. I've changed it so it consistently uses the `mod/customcert:viewreport` permission for downloading certificates.

If you feel like the `mod/customcert:viewreport` permission shouldn't allow a user to download certificates, I can adjust the change so that the `mod/customcert:manage` permission is required to display the download buttons in `classes/report_table.php`.

### DIFF
--- a/view.php
+++ b/view.php
@@ -99,7 +99,7 @@ if ($groupmode = groups_get_activity_groupmode($cm)) {
 }
 
 // Check if we are downloading all certificates.
-if ($downloadall && $canmanage && confirm_sesskey()) {
+if ($downloadall && $canviewreport && confirm_sesskey()) {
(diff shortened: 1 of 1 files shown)


mdjnelson:
Thanks for creating this. I agree consistency would be ideal. I need to think about this a bit.

mdjnelson:
I think this approach is correct. If they can view the report then they should be able to view the certificate, viewing the certificate isnt 'managing'.

mdjnelson:
Cherry-picked, cheers!

Issue #629, opened 2024-07-09
Auto-linking filters move text element position if reference point is center


rhell4:
This issue happens for both "Activity names auto-linking" and "Glossary auto-linking". This is because the html link tags the filters adds around the text are included when getting the string width to calculate the positioning.

Testing instructions:
1. Have a course with the "Activity names auto-linking" filter turned on
2. Add any activity called "Foo"
3. Add a custom certificate activity
4. Add two text elements
    - One "Foo" and the other "Bar"
    - "Reference point location" set to "Center"
    - "Alignment" set to anything other than "Center"
5. Reposition the elements to be in line with each other
6. View the certificate
7. Note that "Foo" is now misaligned and moved to the left

mdjnelson:
Thanks for your work and replication steps, its much appreciated.

PR #630, opened 2024-07-09
Issue #629: strip HTML tags when getting string width

rhell4:
As described in issue #629 the auto-linking filters can mess up text element positions. The actual width should only be calculated from the content without HTML tags. See the issue for testing instructions.

### DIFF
--- a/classes/element_helper.php
+++ b/classes/element_helper.php
@@ -73,7 +73,8 @@ class element_helper {
         $y = $element->get_posy();
         $w = $element->get_width();
         $refpoint = $element->get_refpoint();
-        $actualwidth = $pdf->GetStringWidth($content);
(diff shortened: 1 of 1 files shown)


mdjnelson:
Thanks - looks good to me. The only issue is if people have already adjust the position because of the HTML being taken into account we could ruin their certificates .. need to think how to approach this. Maybe just mention it on the forums, or we could add a hack and only do it for new certificates but that seems yuck. Any thoughts?

rhell4:
I do agree that a hack to only do this for new certificates seems yuck, also that glossary terms and new activities could be created in the future that then effect older certificates which would have this issue. I think mentioning this on the forums/relevant places is a good idea, potentially add it to the changelog too?

PR #631, opened 2024-07-22
Adapt mobile code to ionic 7 

dpalou:
In this PR I'm removing the ionic3 code (the ionic5 app was released on 2021 so it should be safe to remove it).

I also created a new latest template to use the new Ionic syntax for ion-select. The old syntax still works in the 4.4 version of the app (ionic 7), but it will stop working in the 4.5 version (ionic 8) and the ion-select would be wrongly aligned without these changes. To make it easier to review the code, in the first commit I duplicate the template (no changes done in the HTML) and then I apply the changes in a new commit.

Finally, I also removed the core-course-module-description component, it is deprecated and will be removed in a future version of the app. It should have been removed already, but we kept it in the app because some plugins still use it.


dpalou:
Hi Mark! Users can choose whether to update the app or not, we don't force them. Also, our branded apps (and custom apps made by the community) take a while to be updated, so we'll have to keep the ionic 5 template in there for a while.

I just tested this in a 4.1 site and it seems to work fine, so it should be safe to backport, yes. Will you do it or should I create different Pull Requests?

PR #633, opened 2024-07-23
499 Add expiry dates to recipient list, verify

leonstr:
Add a new element, customcertelement_expiry, which represents the certificate's expiry date, 1-5 years from either the award date or the course completion date.  This date is shown in the recipients table shown to the teacher and the verify page when this element is added to a certificate.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -235,5 +235,25 @@ function xmldb_customcert_upgrade($oldversion) {
         upgrade_mod_savepoint(true, 2022112804, 'customcert');
     }
 
+    if ($oldversion < 2022112809) {
+        $elements = $DB->get_records('customcert_elements', ['element' => 'date']);
+
+        foreach ($elements as $element) {
+            $data = json_decode($element->data);
+
+            // If dateitem is between CUSTOMCERT_DATE_EXPIRY_ONE and
+            // CUSTOMCERT_DATE_EXPIRY_FIVE.
+            if ((intval($data->dateitem) <= -8) && (intval($data->dateitem) >= -12)) {
+                $data->startfrom = 'award';
+                $element->data = json_encode($data);
+                $element->element = 'expiry';
+                $DB->update_record('customcert_elements', $element);
+            }
+        }
+
+        // Customcert savepoint reached.
+        upgrade_mod_savepoint(true, 2022112809, 'customcert');
+    }
+
     return true;
 }
--- a/classes/output/verify_certificate_result.php
+++ b/classes/output/verify_certificate_result.php
@@ -62,6 +62,11 @@ class verify_certificate_result implements templatable, renderable {
     public $certificatename;
 
     /**
+     * @var int The certificate's expiry date (optional).
+     */
+    public $expiry;
+
+    /**
      * Constructor.
      *
      * @param \stdClass $result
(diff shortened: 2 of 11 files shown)


leonstr:
This PR is for the MOODLE_401_BRANCH. I also have branches for other supported versions: [499-expirydate-402](https://github.com/leonstr/moodle-mod_customcert/tree/499-expirydate-402), [499-expirydate-403](https://github.com/leonstr/moodle-mod_customcert/tree/499-expirydate-403) and [499-expirydate-404](https://github.com/leonstr/moodle-mod_customcert/tree/499-expirydate-404).

Notes on this proposed change:

1. Because I didn't want to make the core mod_customcert code dependent on the new element I call `class_exists('\customcertelement_expiry\element')` before calling its methods.
2. I added a new element because 1) I wanted it to be possible to start the expiry from the course completion date instead of the award date (only the latter was previously supported), and 2) I wanted to be able to display "Valid for one year" on the certificate, but still have a date appear in the recipients list.
3. It's not possible to use activity completion as the start-from date, this could be added later.
4. The expiry date shown in the recipients list and verify page is wrapped in `<span class="ok|soon|expired">` so admins can change how this is displayed with CSS.
5. Bumped versions because date elements that show an expiry date are migrated in `db/upgrade.php`.
6. Multiple expiry elements are allowed because customcertelement_dates are migrated to this new element, and there may be multiple instances of the old element. (If I wasn't migrating existing elements, I'd only allow a single element per certificate).

db/upgrade.php
```diff
@@ -235,5 +235,25 @@ function xmldb_customcert_upgrade($oldversion) {
...
+    if ($oldversion < 2022112809) {
+        $elements = $DB->get_records('customcert_elements', ['element' => 'date']);
+
+        foreach ($elements as $element) {
+            $data = json_decode($element->data);
+
+            // If dateitem is between CUSTOMCERT_DATE_EXPIRY_ONE and
+            // CUSTOMCERT_DATE_EXPIRY_FIVE.
+            if ((intval($data->dateitem) <= -8) && (intval($data->dateitem) >= -12)) {
+                $data->startfrom = 'award';
```
mdjnelson, line 247: Should this be a string in the language folder?

db/upgrade.php
```diff
@@ -235,5 +235,25 @@ function xmldb_customcert_upgrade($oldversion) {
...
+                $element->data = json_encode($data);
+                $element->element = 'expiry';
```
mdjnelson, line 249: Should this be a string in the language file?

mdjnelson:
This is on my list to look at next. The code looks good from a quick glance, sorry for the delay.

mdjnelson:
Actually, don't worry about it ill resolve them myself. Ill look at putting this in 4.4 first then backporting from there, thanks.

db/upgrade.php
leonstr: `'expiry'`? No, I don't think so, this is the `mdl_customcert_elements.element` value for the new expiry subplugin added in this PR. It's not shown to users. These changes in `db/upgrade.php` are migrating customcertelement_date elements that previously showed an expiry date to the new customcertelement_expiry elements.

mdjnelson: Sorry, if I actually bothered looking at the code in more detail I would have realised that. Using a custom string would break it. Ignore me.

db/upgrade.php
mdjnelson: Sorry, if I actually bothered looking at the code in more detail I would have realised that. Using a custom string would break it. Ignore me.

mdjnelson:
Pushed up the code in 4.4, will see how GHA behaves and backport. Made a few minor adjustments and did some refactoring. Thanks for your hard work, much appreciated.

Issue #387, opened 2020-11-03
Performance issues with email_certificate_task


agrowe:
The `email_certificate_task` scheduled task is set to run every minute but on our instances we are seeing this single task take around 30 seconds to complete with ~27k DB queries on each run.

There are 8 instances over 8 courses (one each) with around 3.5k students total over the courses.

When turning on `$CFG->showcrondebugging = true;` and `$CFG->showcronsql = true;` the main queries look to be the below per each user:

```
SELECT * FROM mdl_course WHERE id = ?
[array (
  0 => '158',
)]

SELECT * FROM mdl_context WHERE contextlevel = ? AND instanceid = ?
[array (
  0 => 30,
  1 => '587',
)]

SELECT ctx.path, ra.roleid, ra.contextid
              FROM mdl_role_assignments ra
              JOIN mdl_context ctx ON ctx.id = ra.contextid
             WHERE ra.userid = ?
[array (
  0 => '587',
)]
```

but with the number of queries, it looks to be roughly 7 queries per student.

Stepping through the task with the SQL output it looks to be related to this line:

```
$cm = get_fast_modinfo($customcert->courseid, $enroluser->id)->instances['customcert'][$customcert->id];
```

`get_fast_modinfo` is then rebuilding the course cache for each user as it loops through `get_enrolled_users`.

I'm not sure if including `$course->cacherev` with `get_fast_modinfo` will fix this or the whole process needs to be reworked, possibly splitting this into two tasks, one for issuing certificates and another for emailing.

The more certificates get added over larger courses, this will cause further issues on each CRON run.

Our mitigation for this now is either to disable this task or change the schedule to nightly instead of each minute.

mdjnelson:
Would it be possible for you to check if `cacherev` does indeed fix the issue? I have very little time these days.

agrowe:
Doesn't look like cacherev has that much on an impact.

I added `$courseobj = get_course($customcert->courseid, false);` to the customcert foreach then updated `get_fast_modinfo` to be `get_fast_modinfo($courseobj, $enroluser->id)`

Before
```
Execute scheduled task: Handles emailing certificates. (mod_customcert\task\email_certificate_task)
... used 26937 dbqueries
... used 20.252206087112 seconds
Scheduled task complete: Handles emailing certificates. (mod_customcert\task\email_certificate_task)
```
After
```
Execute scheduled task: Handles emailing certificates. (mod_customcert\task\email_certificate_task)
[...]

PR #634, opened 2024-07-24
Optimize task email_certificate_task

Claud10R:
Hi!

First of all, thanks for this plugin, it helped me quite a lot in generating custom certificates for my courses. I run a big Moodle installation, counting ~1.000 courses and ~60.000 registered users, and this module is now used in a few selected courses, each of which count thousands of active participants (training courses and language proficiency courses for university students).

In my platform, the task email_certificate_task took up to 15 minutes to run, and it basically ran non-stop, since the default scheduled time is once every minute.

I started digging into the code, and I figured out it was possible to reduce the processing time, mainly by implementing what @alex-rowe suggested in this topic: https://github.com/mdjnelson/moodle-mod_customcert/issues/387 - skipping certificates which are not visible to students. Maybe (not tested) this could be further optimized by changing the first SQL query, adding {course_modules} in the FROM clause and filtering on cm.visible = 1.

I also added another check to skip hidden courses, as suggested by @onadjarj here: https://github.com/mdjnelson/moodle-mod_customcert/pull/610#issuecomment-2242621687. I'm not sure about using courses' end date though, as I'm not sure teachers use it correctly.

Finally, I added a lot of messages in the tasklog, so that it becomes clear what the task has done at every cycle. Log messages require the class attribute VERBOSE to be set to true. Maybe this could also be moved to plugin settings page.


mdjnelson:
Hi Claude, can you rebase this on the latest release? I included a whole lot of changes from other third party devs.

Claud10R:
Hi! I just pulled your changes and everything is good now!
One last optimization could be done by getting the info_module relative to the certificate and using filter_user_list function on it.

```
$infomodule = new \core_availability\info_module($cm);
$filteredusers = $infomodule->filter_user_list($userswithissueview);
```

This way, you could further reduce the list of users to be checked for every certificate, by using the availability restrictions set on the certificate. This function only takes into account those restrictions which implement filter_user_list, so you still need to check $cm->uservisible inside the loop to be 100% sure the user has access to the certificate module.

I also added a verbose mode to print some debug information on the task log, but that's definitely not needed.

mdjnelson:
Do you want to create a separate PR for the `$infomodule->filter_user_list($userswithissueview);` fix?

mdjnelson:
I authored a commit on your behalf @Claud10R with the usage of `filter_user_list`. Closing this now.

Issue #419, opened 2021-04-14
Call list of certificates for a user via API


LaurentLMC:
Hi, 

We are using the certificate module in moodle but our front-end site uses a different CMS and only makes calls to the moodle platform via the API. 

Is there  any way to send a request via the API to know if user X has any certificate available ?

Many thanks in advance for your help !

Lt

mdjnelson:
This would require creating a web service. See https://docs.moodle.org/dev/Web_services. :) I don't have time to create that right now, sorry.

karol-gro:
@mdjnelson I can see in our Moodle web service endpoints which (as I understand) already come from the customcert:
- mod_customcert_delete_issue 
- mod_customcert_get_element_html 
- mod_customcert_save_element 

So does it means that to get the list of issued certificates (for a user or total) it would be sufficient to add a method there?

mdjnelson:
Yes those are the ones that I wrote in for the rearranging of elements page. If you create another one please create a PR I can review. :)

PR #644, opened 2024-10-02
Webservice: List issued certificates

michallohnisky:
Resolves https://github.com/mdjnelson/moodle-mod_customcert/issues/419 .

**Use-case example:**
An external application (CRM) periodically downloads new certificates and associates them with users using their email addresses. The `timecreatedfrom` filter is used to download certificates issued after the last synchronization.


mdjnelson:
Hi,

There are a few things that need addressed.

1. The complaints from the GitHub Actions report.
2. I support Moodle 4.1 upwards so we should be using type hints, return types etc.
3. This does not resolve #419 as that specifies getting certificates for a user, not from a date.
4. Indentation of SQL should be as https://moodledev.io/general/development/policies/codingstyle/sql.
5. There are no checks to see if the user can actually get the list of certificates meaning anyone can get a list of certificates which is a huge security issue.

Please address these and I will take another look.

michallohnisky:
Hi, thanks. I will try my best :-) .

1: Almost done. They disappeared after my push. Can you run them again?
2: Can you give me an example? Because for example `\mod_customcert\external::list_issues_parameters` has a return type in annotations, so it is the same as `\mod_customcert\external::delete_issue_parameters`
3: Done. I added additional filters.
4: I cannot see what is wrong, can you give me an example?
5: Done. I added `mod/customcert:viewallcertificates`

michallohnisky:
1. Done.

michallohnisky:
Are you able to grant me permission to run the checks? If I am able to run them immediately after every push, I think I can add some unit tests as well.

mdjnelson:
You can run the unit tests locally before pushing them. :) https://docs.moodle.org/dev/Writing_PHPUnit_tests

michallohnisky:
I understand :-) , but I would need to set up the whole development stage and Docker containers, and it would take a lot of time for a single unit test.

I squashed the commits.

mdjnelson:
I run the unit tests on my machine and I do not use Docker for Moodle development as I find the Moodle HQ container to be extremely slow. So, its possible to do this without Docker.

PR #655, opened 2024-11-29
Add  new button to return to course

gbarat87:
Hi,
One of our client have requested to add a button in the custom certificate view page.
Here what it looks like now.

image
image

I have added a settings with a default value to not show that button. 

Regards,
Guillaume


mdjnelson:
I have cherry-picked this to 4.4 (which also supports 4.5) and 5.0.

The current patch broke in 4.4 with `The boolean $primary is deprecated and replaced by $type, use single_button::BUTTON_PRIMARY or self::BUTTON_SECONDARY instead`. So, I removed it kept it as a secondary button (there should only be one primary button).

For 5.0 it no longer throws a debugging message as the `single_button` class was moved as the debugging message removed.

The signature for `single_button` in 5.0 is as follows -

```
    /**
     * Constructor
     *
     * @param moodle_url $url
     * @param string $label button text
     * @param string $method get or post submit method
     * @param string $type whether this is a primary button or another type, used for styling
     * @param array $attributes Attributes for the HTML button tag
     */
    public function __construct(
        moodle_url $url,
        $label,
        $method = 'post',
        string $type = self::BUTTON_SECONDARY,
        $attributes = []
    ) {
```    

Again, leaving it as a secondary button.

Issue #659, opened 2025-01-13
Issuing certificates broken


harriholappa:
The following problem was tested on Moodle 4.1.5, mod_customcert 4.1.7 / 2022112814:

1) Make an Assignment activity, with activity completion criteria set to 'Mark as completed'
2) Make a Customcert activity, with an access restriction -> Assignment needs to be completed
3) Set 'email students' as Yes and Set protection: Modify.
4) Add three elements to the certificate: course name, student name and date (I did not configure them further)
5) While logged in with a separate student account, mark the Assignment as completed. Verify with the admin/manager/teappearacher account that the student has completed it in the Activity completion report.
6) As an admin/manager/teacher, navigate to the Recipient view of the certificate.

-> The problem: The student never appears in the list of Recipients unless they click the 'View certificate' button to view the pdf in their browser. 

-> Adjacent problem: The student is also never emailed the certificate, ~~even if they open the pdf and appear on the Recipient list.~~ (EDIT: It seems this was simply because the customcert | useadhoc setting was turned off . The student is emailed the certificate if they click open the certificate pdf, but the automated issuing is not working.)

This has been verified as working in Moodle version 4.1.3 and mod_customcert version 4.1.3 / 2022112806. Has a code change between 4.1.3 and 4.1.7 broken this functionality?

mdjnelson:
This is probably because of https://github.com/mdjnelson/moodle-mod_customcert/issues/646, closing.

mdjnelson:
Ah, thanks for the explanation @anieminen. Are you able to create a PR? :)

anieminen:
I've created a PR with a small modification to the SQL query. That seems to do the trick.

mdjnelson:
I set aside time this weekend to fix a bunch of things with this plugin. Work has been super hectic atm so I apologise for the delay.

mdjnelson:
Pushed up fix, thanks for your patience.

PR #661, opened 2025-01-16
Fix issuing certificate when there are no issues yet (mdjnelson#659)

anieminen:
Fix for issue #659.


mdjnelson:
I've cherry-picked this into supported branches, thanks.

PR #663, opened 2025-02-03
AD-428: feat: added logic to alternate the use of seriescodes instead of default codes

davidherzlos:


davidherzlos:
Sorry, wrong target

PR #667, opened 2025-03-03
Add unique index for code column

### LINKED ISSUE #666: Lookup by code is slow
Looking up certificates is slow. The affects are:

- When creating a new certificate this requires a full table scan to ensure that `code` was not used yet.
- When looking up an existing certificate by `code` this requires a full table scan.

This can be an issue if you are having 100,000+ certificates in your system.

PR coming

fulldecent:
This adds a unique database key to quickly lookup certificates.

Thank you to @Raza403 for contributing his work to this fix.

Fixes #666.

### DIFF
--- a/db/install.xml
+++ b/db/install.xml
@@ -57,6 +57,7 @@
       </KEYS>
       <INDEXES>
         <INDEX NAME="userid-customcertid" UNIQUE="false" FIELDS="userid, customcertid"/>
+        <INDEX NAME="code" UNIQUE="true" FIELDS="code"/>
       </INDEXES>
     </TABLE>
     <TABLE NAME="customcert_pages" COMMENT="Stores each page of a custom cert">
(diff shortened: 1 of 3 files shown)


fulldecent:
@mdjnelson for your review, please

fulldecent:
Hi @mdjnelson could you please approve to run the workflow?

mdjnelson:
Done. The only issue with me merging is the upgrade path. The number needs to be different for each supported version. I will do this myself, no worries but need to set some time aside to go over a few PRs.

fulldecent:
Thank you, much appreciated.

If there's anyway that I can make your job easier next time, please leave some notes in the README or CONTRIBUTING file. 

I have another PR here and planning to work on an additional one after that.

mdjnelson:
Putting this in now. The version bumps are because we dont want to bump all the branches to the same version, as there may be upgrades, say for example, just for 4.2. If someone upgrades to the same version on 4.1 then updates their code to use 4.2 they are going to miss those upgrade steps necessary in 4.2

mdjnelson:
All done, thanks.

Issue #181, opened 2018-04-19
Allow users to load a global template but do not allow them to create their own template within the activity


abias:
We have the use case that users should only be allowed to load a global template but not to create an own template within the activity.

Similar to #137, I propose to add a new capability which allows users to create a custom certificate in the activity.

abias:
On request by @dmitriim, I am adding some background information to this issue:
* Currently, as admin, I can create global templates which users can load in their activity instances.
* Currently, mod/customcert:manage controls multiple aspects of this activity plugin, including the fact that a user can load existing global templates into their activity instances and create own templates within their activity instances.
* As admin, I want to specifically control which users are able to create an own template within an activity instance and I can't do this yet as mod/customcert:manage is some kind of multi-purpose capability.
* As admin, I want to to separate two possible kinds of users:
1. A user wo can load a global template within an activity instance and who can also create his own template within the activity instance. The user will be able to create his own certificates besides the global templates which have been provided by the admin.
2. A user wo can load a global template within an activity instance but who cannot create his own template within the activity instance. The user will only be able to create certificates based on global templates which have been provided by the admin.
* The fact that a user can / cannot modify a loaded template within an activity instance is not covered by this issue. This is handled in #137.

dmitriim:
Hi Alex, 
Sorry, but I'm confused by "users are able to create an own template within an activity instance". Is it possible at all? I can't see this in UI. Am I blind? :)

abias:
Hi Dmitrii,

> Sorry, but I'm confused by "users are able to create an own template within an activity instance". Is it possible at all? I can't see this in UI. Am I blind? :)

I might have phrased this in a mistakable way. What I meant is:
* Login as teacher, go to your course and add an activity instance
* Open the gear icon (in theme_boost) -> Edit certificate
* There, you can load a global template. And you can simply create your own certificate by simply adding elements to your certificate. The second feature is what I meant with "create an own template within an activity instance".

It is clear now?

abias:
This issue is superseded by #227 and will thus be closed.

PR #684, opened 2025-05-11
[WIP] custom_cert added click to download certificate.

Raza403:
@fulldecent please review, this PR is made for [#181](https://github.com/modern-training-solutions/learn.PacificMedicalTraining.com/issues/181).


fulldecent:
@mdjnelson We think `{FIRST_NAME} {LAST_NAME} {COURSE_SHORT_NAME} {DATE}` is a preferable format for students who will download a certificate.

1. Do you agree?; or
2. Do you think we should add a admin-configurable per-template option to set the filename? (This will include a templating language parsing for `{FIRST_NAME}` and other variables.)?; or
3. Are no changes to this in-scope for the project?

rickabitdigital:
> @mdjnelson We think `{FIRST_NAME} {LAST_NAME} {COURSE_SHORT_NAME} {DATE}` is a preferable format for students who will download a certificate.
> 
> 1. Do you agree?; or
> 2. Do you think we should add a admin-configurable per-template option to set the filename? (This will include a templating language parsing for `{FIRST_NAME}` and other variables.)?; or
> 3. Are no changes to this in-scope for the project?

I've had clients requesting specific file-naming so option 2 gets my vote +10. 
They want {GROUP_NAME (which is complicated if users are in more than one group)}{COURSE_SHORT_NAME (so that administrators can group by course)}{FIRST_NAME}{LAST_NAME}.

mdjnelson:
I am happy with option 2. I would build this in the certificate activity settings rather than an admin setting just to clarify.

fulldecent:
Hi @Raza403 I updated the specifications here https://github.com/modern-training-solutions/learn.PacificMedicalTraining.com/issues/181

mdjnelson:
Are you able to address the issues listed above? Thanks.

Raza403:
@fulldecent please review, added settings.
Screenshot 2025-06-14 at 11 06 03 PM

fulldecent:
Hi @Raza403 above we had discussed allowing the admin to type in a string. "Option 2". And this string could include text like `{FIRST_NAME}` that would be replaced with the appropriate value at the time of generating the certificate. 

I did not see this implemented in your photo. And you please able to implement this and show in a new picture?

Raza403:
Got it, so two things here.
1. String could include text like {FIRST_NAME} that would be replaced with the appropriate value at the time of generating the certificate.
2. Build this in the certificate activity settings rather than an admin setting.

fulldecent:
@Raza403 yes you got it perfect

Raza403:
@fulldecent, please review the updated code --- screenshots attached below.

📄 **Custom name example:**\
`Ahmed_Raza_DS-1_2025-06-15_Test_group.pdf`

📄 **Pattern string used:**\
`{FIRST NAME} {LAST NAME} {COURSE SHORT NAME} {DATE} {GROUP}`

👉 To achieve this, I added an array of substitution values directly in `template.php`, as these placeholders are **not available by default in Moodle**:

```php
$values = [
    '{FIRST NAME}' => $user->firstname ?? '',
    '{LAST NAME}' => $user->lastname ?? '',
    '{COURSE SHORT NAME}' => $course ? $course->shortname : '',
    '{COURSE FULL NAME}' => $course ? $course->fullname : '',
    '{DATE}' => $issuedate,
    '{GROUP}' => $groupname ?? ''
];
```

We then apply these replacements using `strtr()`.

Screenshot 2025-06-15 at 11 52 35 PM
Screenshot 2025-06-15 at 11 54 18 PM

Raza403:
@fulldecent this is good to review, can you please review this PR and see my comment above^^^.

mdjnelson:
Please rebase on the latest MOODLE_404_STABLE branch as I have updated the latest Moodle Plugin CI version,

Raza403:
@mdjnelson rebased.

fulldecent (approved):

mdjnelson:
Looking now. :)

lang/en/customcert.php
```diff
@@ -71,10 +76,14 @@
 $string['deleteissuedcertificates'] = 'Delete issued certificates';
 $string['deletepageconfirm'] = 'Are you sure you want to delete this certificate page?';
 $string['deletetemplateconfirm'] = 'Are you sure you want to delete this certificate template?';
+$string['deliveryoption'] = 'Delivery option';
+$string['deliveryoption_help'] = 'Choose how the certificate should be delivered to users.';
 $string['deliveryoptiondownload'] = 'Send to the browser and force a file download';
 $string['deliveryoptioninline'] = 'Send the file inline to the browser';
 $string['deliveryoptions'] = 'Delivery options';
 $string['description'] = 'Description';
+$string['digits-with-hyphens'] = '0123-4567-8901 (Digits with hyphens)';
```
mdjnelson, line 85: This is never used.

db/upgrade.php
```diff
@@ -316,5 +316,23 @@ function xmldb_customcert_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2024042210, 'mod', 'customcert');
     }
 
+    if ($oldversion < 2025062100) {
```
mdjnelson, line 319: This should be the value of the version for version.php before it was bumped. The reason is if we use the current date and then do an upgrade, then minor versions would have the same version as the `main` branch. This can cause issues because we may have introduced an upgrade step 2 weeks ago on just `main` for a new feature then when people upgrade to the `main` they miss the upgrade as the date they have is past that.

mdjnelson:
What would be great is putting the logic for generating the file name and putting this in a separate method we can then unit test.

mdjnelson:
https://github.com/mdjnelson/moodle-mod_customcert/pull/702/
https://github.com/mdjnelson/moodle-mod_customcert/pull/703/

Please take a look at the above for future reference on things to check before creating a PR. The big things here were the version issue as described in one of my PR comments and the help string not reflecting what was actually in the code (this functionality didn't work when I followed the _help string; I then looked at code to see what was expected).

mdjnelson:
I created https://github.com/mdjnelson/moodle-mod_customcert/issues/704. If you could please address this that would be excellent.

mdjnelson:
Ah. Also forgot about backup/restore if there are any new fields.

mdjnelson:
Ah, also yes the upgrade code is necessary however this is not run on fresh installs so you need to update `install.xml` as well.

fulldecent:
Got it. @Raza403 is it possible to have our change only set the file name without doing force download?

fulldecent:
Got it OK thank you. Let's rollback force download because Honey doesn't like it.

Raza403:
Done.

Issue #693, opened 2025-06-16
Fix notifications when the instance of the activity is on site home


andrewhancox:
If you create an instance of this activity on the site home course then, because the site home course has no category, the query that generates email notifications never picks it up.

mdjnelson:
Thanks! Pulled in. :)

PR #695, opened 2025-06-16

andrewhancox:


mdjnelson:
Closing as 4.2 is no longer supported. :)

PR #696, opened 2025-06-16

andrewhancox:


mdjnelson:
Closing as Moodle 4.3 is no longer supported.

PR #697, opened 2025-06-16

andrewhancox:


andrewhancox:
That course is redundant, I've removed it and force pushed and amended commit to both this branch and MOODLE_500_STABLE

PR #698, opened 2025-06-16

andrewhancox:


mdjnelson:
Fixed the issues that GHA was complaining about. Done.

PR #706, opened 2025-07-29
test: create/delete unit test

fulldecent:


Raza403:
You're right — after investigating, I found that the extra commits are coming from our base fork.

It looks like we previously created PRs for those changes, and @mdjnelson may have incorporated them directly into the upstream branch without merging our PRs. As a result, our fork's MOODLE_404_STABLE now contains all those commits, even though they were already merged separately upstream. This causes them to appear in new PRs as part of the diff.

mdjnelson:
You can always force push with git instead of creating another PR. :)

PR #728, opened 2025-11-20
Certificatelink

### LINKED ISSUE #724: Text in email body after link
In HTML emails the link is always the last element. In our Moodle instance we would like to modify the language string for the email body to end in:

> Best regards
> Your Moodle team

But the link will always come after this text which doesn't look good.

Marvin-Kirch:
Fixes #724 

I also added two more commits, one of which standardizes the structure of the HTML version with the plain text version, and one that takes the user's language into account when sending emails.


mdjnelson:
Done, thanks!

Issue #720, opened 2025-09-17
Support an individual fullname


mwreichel:
Hi,
at our university, we use the custom certificate in teaching and for internal and external training courses.
On our Moodle installations we adapted the general settings for fullname and and the alternative name in the user policies. This information is used as well in the plugin for the elements ‘studentname’ and ‘teachername’.
We were asked to support a different form of the fullname with certificates because there it is another purpose than usual. This not possible at the moment.

Therefore we plan to prepare the following changes and would like to submit a PR afterwards:

**Description**
* add an admin setting in the plugin to define an individual fullname in the certificate
* by default, the field of the admin setting is empty -> behaviour: the fullnamedisplay of the core setting is used (as now)
* the new field can be filled with some fields from the user table (similar to the core feature) -> behaviour: this value is used in the certificate instead the fullnamedisplay of the core setting
* the elements ‘studentname’ and ‘teachername’ have not to be adapted by users after this change, -> behaviour: if there's an individual fullname defined it will be applied from that moment on

If you would accept this PR after we have finished work on it, we would be delighted.

All the best,
Martin

mdjnelson:
Sorry for the delayed response. Please, yes, create a PR and I will happily review it. :)

mwreichel:
Hi @mdjnelson,
great thank you. We should have finished work on the PR soon.

Cheers,
Martin

PR #736, opened 2025-12-19
Support an individual fullname #720

stellargela:
add #720

### DIFF
--- a/classes/fullname_certificate.php
+++ b/classes/fullname_certificate.php
@@ -1,0 +1,143 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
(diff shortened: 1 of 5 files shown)


mwreichel:
Hi @mdjnelson,
thank you for your feedback to the fixed version of @stellargela.

So would it be ok for you if we implement these admin settings to the elements in the same way instead?

Our approach with an admin setting in the core plugin was to define a global naming scheme. In our case the defined scheme should be used automatically by any element plugin and editing users shouldn't be able to define the scheme on their own.

Thanks in advance,
Martin

mdjnelson:
Thanks for explaining the use case. I understand why you want a site-wide certificate name format, and I’m fine with that being configured by an administrator rather than by each editing teacher.

My concern is where the logic lives. This behaviour is specific to the `studentname` and `teachername` certificate elements, so I’d prefer the settings and formatting logic to live inside the relevant element plugin(s), not in the core `mod_customcert` plugin. The core plugin should not need to know that these particular elements have a custom fullname policy, nor should it contain helper `classes/settings` that only those elements use.

mod_customcert already supports element plugins loading their own settings: after the main customcert settings, it creates the customcertelements category and calls each customcert element plugin’s load_settings() method. So an admin-level, globally enforced setting can still be implemented, but it should be owned by the element plugin rather than by mod_customcert itself. This keeps the separation clean: core provides the certificate/element framework, while each element owns its own rendering policy.

In this PR, the new `mod_customcert\fullname_certificate` class and the `customcert/fullname` + `customcert/fullnameformat` settings are added to the core plugin, but the settings are only consumed by the `studentname` and `teachername` elements. That creates coupling between core and specific element behaviour. If another name-related element wanted similar behaviour later, we’d either have to keep expanding core or make the global setting mean different things for different elements.

Could you move this into the relevant element plugin(s) instead? For example, the element plugin could expose the admin setting, contain the formatting helper, and use it consistently in both PDF rendering and HTML preview. The setting can still be global/admin-controlled, so editing users do not need to choose the format per certificate element.

Thanks for your patience and hard work! :)

PR #759, opened 2026-02-01
Use issue userid when generating downloads and add regression test

mdjnelson:
Fixes a bug.


mdjnelson:
The id in this case is the user id.

PR #788, opened 2026-04-16
feat: activity completion when certificate is issued via email

nersesov:
Closes #645

Adds a new custom completion rule "Receive a certificate" (`completionissued`) that automatically marks the activity complete when a certificate is emailed to the student via the scheduled task.

**Changes:**
- `classes/completion/custom_completion.php` — Moodle 4.x `activity_custom_completion` API implementation
- `classes/task/completion_backfill_task.php` — adhoc task to retroactively set completion for users who already received a certificate
- `classes/task/issue_certificates_task.php` — triggers completion after certificate issuance
- `db/install.xml`, `db/upgrade.php` — adds `completionissued` field to `mdl_customcert`
- `lib.php` — declares `FEATURE_COMPLETION_HAS_RULES`
- `mod_form.php` — adds checkbox to activity settings, disabled when "Email students" is off
- `lang/en/customcert.php` — new strings

Happy to provide branches for MOODLE_500_STABLE / main if needed.

PR #792, opened 2026-04-17
feat: activity completion on certificate issue

### LINKED ISSUE #645: Add activity completion rule when a certificate email is sent
The automatic completion settings for this activity only give the option to be based on the student having viewed the activity, but we've seen that many students don't view the activity because they also received the certificate by email.

Because of that, many students appear to haven't completed the activity in the activity completion reports even if they've already received the certificate.

Having that into account, it would be great if there were another completion requirement option based on having received the certificate.

nersesov:
Closes #645

Re-opening #788 for main as requested — rebased and adapted to the new architecture introduced in the recent refactor.

Adds a new custom completion rule "Receive a certificate" (completionissued) that automatically marks the activity complete when a certificate is emailed to the student via the scheduled task.

Changes:

[custom_completion.php](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — activity_custom_completion API implementation
[completion_backfill_task.php](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — adhoc task to retroactively set completion for users who already received a certificate
[certificate_issuer_service.php](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — triggers completion after certificate issuance, adapted to the new service layer (certificate_issuer_service / certificate_repository)
[certificate_repository.php](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — includes completionissued in the SQL SELECT
[install.xml](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), [upgrade.php](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — adds completionissued field to mdl_customcert (version 2025122801)
[lib.php](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — removes FEATURE_GROUPS (already absent in main)
[mod_form.php](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — adds checkbox to activity settings, disabled when "Email students" is off
[customcert.php](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — new strings: completionissued, completionissued_help, completionissuedemailerror, taskcompletionbackfill


mdjnelson:
Thanks for the contribution. I think the feature is useful, but I do not think this should be merged in its current form.

My main concern is that the patch mixes together three slightly different concepts:

1. Certificate is **issued**.
2. Certificate is **emailed**.
3. Student **receives certificate by email**.

At the moment the implementation appears to use `customcert_issues.emailed = 1` as the real source of truth, but the field/rule naming is `completionissued`, and the commit/title/docs read more like "complete on issue". I think this needs to be made consistent before merging, because the current version makes the intended behaviour hard to reason about.

A few specific issues I noticed:

1. **Completion is updated at the wrong point for an email-based rule**
In `certificate_issuer_service`, completion is marked complete immediately after `issue_if_needed()`, before the email send/queue path runs. If this rule is meant to mean “certificate emailed”, then the completion update is happening too early.

2. **Custom_completion and manual completion updates are overlapping**
The patch adds a `custom_completion` rule, but also manually forces completion via `completion_info->update_state(... COMPLETION_COMPLETE ...)`. That creates two sources of truth. I think this should either:

-  Rely on the custom completion rule and trigger reevaluation when state changes, or ..
-  use imperative completion updates only, but not both.

3. **The rule evaluation does not clearly honour `completionissued`**
 In `custom_completion::get_state()`, the logic checks email settings, but it does not obviously use the instance’s `completionissued` flag as the rule toggle. That makes the runtime behaviour harder to follow.

4. **Upgrade/backfill is too aggressive**
 The adhoc backfill task retroactively sets `completionissued = 1` for existing activities. I do not think we should silently opt existing activities into a new completion rule during upgrade. Completion settings are user-facing behaviour, so this feels too invasive. My preference would be:

- Add the field defaulting to `0`.
- Leave existing activities unchanged.
 - only backfill completion for activities that are explicitly opted into the new rule.

5. **Backfill criteria are inconsistent with runtime behaviour**
 The backfill only selects instances with `emailstudents = 1`, while the runtime logic also considers the global config. Those two paths should be aligned.

6. **`FEATURE_GROUPS` removal looks unrelated**
 The removal of `FEATURE_GROUPS` in `lib.php` does not seem related to this feature and should not be bundled into this patch unless there is a separate reason for it.

Overall, I think the feature would be much easier to review and maintain if it were reworked around one precise rule:

- either **complete when certificate is issued**
- or **complete when certificate email is sent**

and then use that meaning consistently in:

- field names
- help text
- custom completion logic
- service hooks
- migration/backfill behaviour

I’d be happy to review a revised version once the semantics are tightened up and the upgrade behaviour is made less invasive.

## Summary

1. Pick one precise completion meaning and use it consistently everywhere: either complete when certificate is issued or complete when certificate email is sent.
2. Rename the setting, rule, and language strings to match that meaning. completionissued is misleading if the actual trigger is email-related.
3. Move the completion update to the correct lifecycle point for the chosen rule. If it is email-based, completion should not be marked immediately after issue_if_needed().
4. Avoid having two sources of truth. Either rely on custom_completion plus reevaluation, or use imperative completion updates, but not both.
5. Make the custom completion evaluation explicitly respect the instance setting that enables the rule.
6. Remove the automatic upgrade opt-in/backfill behaviour, or at minimum restrict any backfill to activities explicitly opted into the new rule.
7. Align backfill criteria with runtime behaviour so both paths use the same logic for per-activity and global email settings.
8. Drop the unrelated FEATURE_GROUPS change from this patch unless there is a separate reason for including it.

Thanks! :)

mdjnelson:
I went through and did all these changes as they weren't too difficult. Please take a look.

nersesov:
Thank you for taking the time to rework this - the changes are clean and correct.
The rename to `completionemailed`, the single source of truth via `custom_completion`,
moving the trigger to `certificate_email_service`, dropping the backfill, and adding
`get_coursemodule_info()` + `FEATURE_COMPLETION_HAS_RULES` are all exactly right.
The test coverage is thorough too.

Happy to see this merged as-is. Thanks again!

nersesov:
Hi @mdjnelson, just a gentle ping on this one. As far as I can tell the PR is in good shape after your rework, and I'm happy with the current state. Let me know if there's anything else you'd like me to address before merging, or if you're waiting on something on your end. Thanks again for taking the time to clean it up!

mdjnelson:
Hey @nersesov. Sorry, just been very busy on rewriting this plugin and getting rid of technical debt. I will most likely only put this into `main` which will become 5.3 when that is released so there is still some time before I have to merge but it's on my list.

mdjnelson:
Closing. Will be doing a PR from my own branch back into main. I can't make changes to https://github.com/nersesov/moodle-mod_customcert/tree/feat/completion-on-cert-issue-main as it is your repository.

mdjnelson:
My PR can be found @ https://github.com/mdjnelson/moodle-mod_customcert/pull/903.

nersesov:
Thank you very much. I really appreciate it 👍🏻

PR #794, opened 2026-05-12
Fix: Remove use() from DI closures for prod compilation compatibility

Jayce0808:
php-di's container compiler cannot compile closures that capture variables via the use() keyword. This refactors the DI definitions to register template_logger as a singleton service and inject it as a parameter instead, making the closures compilable in production mode.


mdjnelson:
Thanks! I have merged this. :)

PR #796, opened 2026-05-12
Fixes #795 - Cannot compile closures which import variables using the…

### LINKED ISSUE #795: Uncaught exception 'DI\Definition\Exception\InvalidDefinition'
Hi, it looks like there will be an issue with `public/mod/customcert/classes/local/hook_callbacks.php` in production codebases. `public/lib/php-di/php-di/src/Compiler/Compiler.php:389` does not allow closures which import variables using the `use` keyword but these have been introduced after the refactor. 

https://github.com/mdjnelson/moodle-mod_customcert/pull/794

willianmano:
… use keyword

### DIFF
--- a/classes/local/hook_callbacks.php
+++ b/classes/local/hook_callbacks.php
@@ -51,12 +51,10 @@ class hook_callbacks {
      * @param di_configuration $config The DI configuration instance to register definitions in.
      */
     public static function di_configuration(di_configuration $config): void {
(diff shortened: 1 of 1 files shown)


mdjnelson:
Thanks for the PR but https://github.com/mdjnelson/moodle-mod_customcert/pull/794 just beat you to it. :)

PR #801, opened 2026-05-26
Add Moodle Playground demo (blueprint + README badge)

erseco:
## About this PR — Moodle Playground

[Moodle Playground](https://moodle-playground.com) ([source](https://github.com/ateeducacion/moodle-playground)) runs a full Moodle site entirely in the browser (PHP + SQLite via WebAssembly). A `blueprint.json` provisions plugins and sample content so anyone can try a plugin with one click.

## Current scope (simplified)

This PR now only adds:

1. **`blueprint.json`** — Moodle demo with `mod_customcert` installed, demo course `CCERTDEMO01`, a preloaded certificate with sample text elements, ready to issue/download a PDF. The blueprint owns the Moodle version for this branch (`main` → 5.3).
2. **`README.md`** — short “Try in Moodle Playground” badge pointing at **this repository’s** `main/blueprint.json` so it works as soon as this is merged.

The earlier GitHub Action / PR-preview workflow has been **removed** to keep the contribution small and avoid `pull-requests: write` and related trade-offs. See the latest comment for context.

## Test plan

- [ ] Open the Playground demo from the comment below and confirm Moodle boots with the plugin installed.
- [ ] Open the demo certificate activity, issue/download a PDF, or edit the template.
- [ ] Confirm this PR does not add any GitHub workflow under `.github/workflows/`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,17 @@
 
 This activity allows dynamic generation of PDF certificates with full customisation in your browser.
 
+
+## Try in Moodle Playground
+
+Click the badge below to open this plugin instantly in
+[Moodle Playground](https://moodle-playground.com) — a full Moodle site
+running in the browser, with no local install. The demo includes a course
+with a preloaded certificate activity so you can edit the template, issue
+and download a PDF immediately.
+
+<a href="https://moodle-playground.com/?blueprint-url=https://raw.githubusercontent.com/mdjnelson/moodle-mod_customcert/refs/heads/main/blueprint.json" target="_blank" rel="noopener"><img src="https://raw.githubusercontent.com/ateeducacion/action-moodle-playground-pr-preview/refs/heads/main/assets/playground-preview-button.svg" alt="Preview in Moodle Playground" width="200"></a>
+
 ## Requirements
 
 - A supported Moodle version (use the plugin release/branch that matches your Moodle version).
--- a/blueprint.json
+++ b/blueprint.json
@@ -1,0 +1,77 @@
+{
+  "$schema": "https://raw.githubusercontent.com/ateeducacion/moodle-playground/refs/heads/main/assets/blueprints/blueprint-schema.json",
+  "preferredVersions": {
+    "php": "8.3",
+    "moodle": "5.3"
+  },
+  "landingPage": "/course/view.php?id=2",
+  "constants": {
+    "ADMIN_USER": "admin",
+    "ADMIN_PASS": "test",
+    "ADMIN_EMAIL": "admin@example.com"
+  },
+  "steps": [
+    {
+      "step": "installMoodle",
+      "options": {
+        "adminUser": "{{ADMIN_USER}}",
+        "adminPass": "{{ADMIN_PASS}}",
+        "adminEmail": "{{ADMIN_EMAIL}}",
+        "siteName": "Custom Certificate Activity Demo",
+        "locale": "en",
+        "timezone": "Europe/Madrid"
+      }
+    },
+    {
+      "step": "login",
+      "username": "{{ADMIN_USER}}"
+    },
+    {
+      "step": "installMoodlePlugin",
+      "pluginType": "mod",
+      "pluginName": "customcert",
+      "url": "https://github.com/mdjnelson/moodle-mod_customcert/archive/refs/heads/main.zip"
+    },
+    {
+      "step": "createCategory",
+      "name": "Test Courses"
+    },
+    {
+      "step": "createCourse",
+      "fullname": "Custom Certificate Activity Demo Course",
+      "shortname": "CCERTDEMO01",
+      "category": "Test Courses",
+      "summary": "Demo course for the mod_customcert activity module. A sample certificate activity with a default template is preloaded so you can immediately open the template editor and design a certificate, or issue and download a PDF.",
+      "format": "topics",
+      "numsections": 3
+    },
+    {
+      "step": "createUser",
+      "username": "student",
+      "password": "test",
+      "email": "student@example.com",
+      "firstname": "Demo",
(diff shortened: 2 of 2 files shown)


mdjnelson:
I am just about to head to work and I would have to look at this in more detail. :)

Also, yep if you could do a separate PR for the component caching issue that would be great to look at in isolation and keep separate from this new feature.

erseco:
Quick update on the `element_bootstrap` defensive change I mentioned in the PR description:

After digging into the root cause, the bug is **not** in `mod_customcert` — the plugin's `db/subplugins.json` is correct and a normal Moodle install handles it fine. The failure mode is specific to Moodle Playground: after `installMoodlePlugin` installs the plugin into the WASM filesystem, its `core_component::playground_refresh_installed_plugin_cache()` patch was refreshing the parent plugin's classmap/filemap but **not reading the freshly installed plugin's `db/subplugins.json`** — so the autoloader never learned about the `customcertelement_*` subplugin classes and `class_exists()` returned false inside `element_registry::register()`.

That issue affects every Moodle plugin with subplugins installed via Playground (mod_quiz / `quizaccess_*`, mod_assign / `assignsubmission_*`, mod_workshop / `workshopform_*`, mod_qbank, …), so the correct place to fix it is playground itself, not each plugin.

Fixed upstream and already merged: https://github.com/ateeducacion/moodle-playground/pull/108 — `playground_refresh_installed_plugin_cache()` now processes `db/subplugins.json` (both the modern `plugintypes` and legacy `subplugintypes` formats), registers the subplugin types in `self::$plugintypes` / `self::$subplugins`, and refreshes each discovered subplugin's classmap before writing the cache.

I've therefore **dropped the `element_bootstrap.php` patch from this PR** and force-pushed — the branch now only contains the Playground tooling (`blueprint.json`, the PR-preview workflow, and the README section). No `mod_customcert` PHP file is modified anymore, and the live preview already works end-to-end with the upstream playground fix.

Sorry for the noise about a separate "component caching" PR — it's no longer needed. 🙂

mdjnelson:
Thanks for the PR. I like the idea of having a one-click Moodle Playground preview for reviewers.

I think a few things need addressing before merge:

1. The playground is hardcoded to Moodle 5.2 in the workflow, README and blueprint. Could you update the playground configuration/documentation so the Moodle version is not unnecessarily fixed to 5.2? Ideally it should either derive the appropriate version from the target branch, or be documented/maintained per branch. I intend to include this in the 5.2 branch as well, but it should also remain useful going forward. Otherwise the generated preview may fail to install the plugin, or give reviewers a misleading test environment when used on branches targeting a different Moodle version. For example, `main` now requires Moodle 5.3 dev.

2. GitHub is warning that the added files contain hidden/bidirectional Unicode characters. Could you please remove those and re-save the files as plain UTF-8? I’d prefer not to merge workflow/JSON files with invisible characters.

3. The workflow is restricted to same-repo PRs. That may be intentional from a permissions/security point of view, but it means forked contributor PRs will not get previews. Is that the intended trade-off? If so, the README wording should stay explicit about “same-repo” previews.

4. Minor security/maintenance preference: could you avoid using `ateeducacion/action-moodle-playground-pr-preview@main`? Since this workflow has `pull-requests: write`, I’d prefer the action to be pinned to a release tag, or ideally a full commit SHA, rather than a mutable branch. Not a blocker if there is not currently a stable release/tag for the action, but I think it would be safer long-term.

Thanks! :)

erseco:
Hi, and thank you so much for taking the time to review this.

I've updated the PR to address every point you raised:

- **Removed the unnecessary Moodle 5.2 hardcoding** from the workflow and the README. The blueprint now owns the Moodle version for the target branch.
- **Updated the blueprint to Moodle 5.3** to match the version required by the target branch (`main`).
- **Kept previews intentionally limited to same-repository pull requests.** This is a deliberate security decision: publishing the preview requires write access through `GITHUB_TOKEN`, and forked PRs should not receive write-capable credentials.
- **Documented that limitation explicitly in the README**, so the behaviour is clear to other contributors.
- **Removed any hidden Unicode characters** from the added files — they are now plain UTF-8 with LF line endings (verified, and `git diff --check` is clean).
- **Replaced the mutable `@main` action reference** with the full immutable commit SHA for the current `v1` release (`bfa63e4863a8cf9d6ac9394bd83f8669000dfd51`).

I've kept the PR focused on these review fixes only. Thanks again for the guidance, please let me know if you'd like any further changes!

erseco:
Hi Mark — following up on your review.

I've simplified this PR quite a bit.

Your feedback (especially around `pull-requests: write`, same-repo-only previews, and pinning the action) made me reconsider how invasive the GitHub Action was for a plugin repository. Rather than keep iterating on the workflow, I've **removed it entirely**.

What remains is a much smaller contribution:

- `blueprint.json` — one-click demo (Moodle version owned by the blueprint; currently aligned with `main` / 5.3, as you requested)
- `README.md` — short “Try in Moodle Playground” badge (points at this repo so it works as soon as this is merged)

No workflow, no extra permissions, no PR auto-comments. That also avoids the same-repo vs fork trade-off and the need to pin an external action.

You can try the demo from this branch here:

<a href="https://moodle-playground.com/?blueprint-url=https://raw.githubusercontent.com/erseco/moodle-mod_customcert/refs/heads/feature/moodle-playground-preview/blueprint.json" target="_blank" rel="noopener">Preview in Moodle Playground</a>

Sorry for the extra back-and-forth — happy to adjust the blueprint/README further if you'd like anything different. Thanks again for the careful review.

mdjnelson:
I created a new PR as I added new commits, see https://github.com/mdjnelson/moodle-mod_customcert/pull/843.

Issue #804, opened 2026-06-05
get_instance_data(): $instanceid must be of type int, string given with custom course fields


leonstr:
### Steps to reproduce issue

1. Add a custom course field in Moodle.
2. Create a course populating the above field for the course.
3. Add a mod_customcert instace for the course.
4. Edit the certificate adding a _Course field_ element.
5. Set the element's _Course field_ setting to the custom course field and click _Save changes_.
6. Click Save changes and preview.

### Expected results

The certificate preview is displayed showing the custom course field value.

### Actual results

_Exception - core_customfield\handler::get_instance_data(): Argument #1 ($instanceid) must be of type int, string given, called in [dirroot]/mod/customcert/element/coursefield/classes/element.php on line 189_

### Environment

Moodle 5.2.1 (Build: 20260608), mod_customcert 2026042001 (commit 426b741).

mdjnelson:
Done. Thanks!

PR #805, opened 2026-06-05
fix $instanceid must be type int with coursefield

leonstr:
Proposed fix for https://github.com/mdjnelson/moodle-mod_customcert/issues/804


leonstr:
Actually maybe this needs PHPUnit test coverage?

mdjnelson:
Added unit test. :) Now in the code. Thanks!

PR #806, opened 2026-06-08
get_mod_grade_info needs integers

tknaller:
get_mod_grade_info needs integers since commit 19a43fe


mdjnelson:
Thanks a lot for this. I found some other places as well. Am fixing them now and including this as well. :)

PR #842, opened 2026-07-14
Fix N+1 query in issue_certificates_task issuance loop

jonycheung:
## Problem

`certificate_issuer_service::process_email_issuance_run()` iterates every eligible ("candidate") user for a certificate and calls `issue_if_needed()`, which unconditionally runs `issue_repository::find_by_user_certificate()` — a `$DB->get_record()` call against `customcert_issues` — before deciding whether to create a new issue:

```php
public function issue_if_needed(int $customcertid, int $userid): ?object {
    $issue = $this->issues->find_by_user_certificate($customcertid, $userid);
    ...
}
```

That's one DB round-trip per candidate user, per certificate — a classic N+1. Confirmed in production on a course with **14,114 enrolled users**, where the `issue_certificates_task` scheduled task took **25–78+ minutes per invocation** and drove **sustained 100%+ database CPU for hours**.

This is a distinct, but related, problem to #387, which covers the same class of N+1 issue via `get_fast_modinfo()` calls in the candidate-filtering step (marked duplicate, still unresolved). This PR does not touch `get_fast_modinfo()` — it fixes the `customcert_issues` query specifically.

## Fix

Bulk-prefetch every `customcert_issues` row for the certificate **once**, before the per-user loop, via a new `issue_repository::list_by_certificate_keyed_by_userid()`:

```php
$existingissues = $DB->get_records('customcert_issues',
    ['customcertid' => $customcert->id], '', 'userid, id, emailed');
```

`issue_if_needed()` now takes this prefetched map as an optional parameter. When a user is already in the map, no query is made at all — this collapses the dominant cost (one query per already-issued candidate) into a single bulk read per certificate.

## Race-condition handling

`customcert_issues` has **no unique constraint** on `(userid, customcertid)` (only a non-unique index — see `db/install.xml`), so the existing check-then-insert in `issue_if_needed()` was already a (small, millisecond-scale) race window, not something newly introduced here.

To avoid widening that window: the prefetched map is only used to **skip** work for users it says are already issued — that's safe, since a stale "yes, exists" read can't cause a duplicate. For users the map says are new, we deliberately do **not** trust the (potentially stale) prefetch for the write path. Instead we fall through to a fresh, tight `find_by_user_certificate()` check immediately before `create()`, exactly as before. So genuinely-new issuance keeps its original (already-accepted) race characteristics, while the large majority of already-issued users on big courses skip the query entirely.

## Tests

Added to `tests/email_certificate_task_test.php`:

- `test_issue_if_needed_skips_query_for_prefetched_existing_issue` — asserts `$DB->perf_get_reads()` is unchanged when the issue is already in the prefetched map.
- `test_issue_if_needed_issues_new_user_not_in_prefetched_map` — a user absent from the map is still issued correctly via the fresh check-then-insert path.
- `test_process_run_creates_no_duplicate_issues_with_bulk_prefetch` — a mix of a pre-issued-but-unemailed user (simulating a prior interrupted run) and a genuinely new user produces exactly one row each, with correct `emailed` flags, and re-running the task creates no duplicates.
- `test_process_run_read_count_does_not_scale_with_already_issued_candidates` — compares marginal DB reads between a small and large batch of already-issued-but-unemailed candidates and asserts the growth is far below 1:1 with candidate count (pre-fix, it was ~1:1).

I wasn't able to run the suite in this environment (no local PHP/Moodle install available to me), so please run CI/PHPUnit on this PR before merging.

## Notes

- Targeted at `main` per the repository's default branch (no `CONTRIBUTING.md` was found specifying otherwise).
- `issue_if_needed()`'s new third parameter is optional and defaults to `[]`, so existing callers/tests are unaffected.


mdjnelson:
Hi @jonycheung, thanks for this! Are you able to address the issues spotted by GHA? :)

jonycheung:
> Hi @jonycheung, thanks for this! Are you able to address the issues spotted by GHA? :)

[mdjnelson](https://github.com/mdjnelson) Updated the branch. Could you run GHA for me

mdjnelson:
Running now. However, I still see "This branch cannot be rebased due to conflicts" so a rebase + conflict resolution will be needed.

cameron1729:
Hey @mdjnelson I don't think this should merge. There are some blockers here:

- The prefetched email flag can become stale during a long cron run, so like an adhoc task could send a certificate then queue it again because it still sees `0`
- The new `get_records` result is indexed by user ID, but duplicate issue rows are legit. As is, this can overwrite rows in the result and emit debugging notices
- The N+1 thing the patch claims to address remains for every cache miss. Also, the task now loads every issue up front, the performance test only covers cache hits and starts measuring after the prefetch (so it misses this)

jonycheung:
Thanks @cameron1729 — all three were right, fixed in d547867.

- **Stale email flag.** `send_issue()` now re-reads the flag when it sends and returns early if set — no extra query, `get_user_for_issue()` already selects it. This one predates the prefetch: a backed-up queue means run 2 re-queues before task 1 has run. To be precise, duplicate tasks can still be *queued*, they just no-op now. Deduping at queue time costs a query per candidate, and concurrent sends would need mark-before-send, trading double-sends for lost emails on retry. Happy to go either way if you'd prefer.
- **Duplicate rows.** Reproduced exactly as you described. The map is built from a recordset now rather than keyed by the DML layer — earliest row per user wins, already-emailed takes precedence.
- **N+1 on cache miss.** The map is authoritative for the users it loaded, so misses go straight to creation. Residual is a duplicate row if it goes stale, which the schema already permits and check-then-insert couldn't prevent either. Prefetch is scoped to the candidates (chunked `IN`) rather than every issue.
- **Test.** Good catch — and my first fix was still weak: I measured the baseline *through* `issue_if_needed()`, so when I reintroduced the N+1 to check the test had teeth, it inflated the baseline and still passed. Baseline now comes from `issue_repository::create()`, counting starts before the prefetch, and misses are covered; with the N+1 put back it fails `121 !== 81`. Each new test confirmed to fail without its fix.

431 tests pass on Moodle main / MariaDB 10.11; phpcs and PHPDoc checker clean.

mdjnelson:
@jonycheung can you put this in one commit and address the issues GHA found. Thanks!

mdjnelson:
Ok, had time to actually look at it and not just wait for the GHA jobs to go green before seeing the changes.

I think this is heading in the right direction and the changes since @cameron1729 review make sense.

I am still a bit concerned about the stale prefetch though. The comment says that a stale map can cause a duplicate issue row but can no longer cause a duplicate email because `send_issue()` re-reads the email flags. I don't think that is quite true.

For example, the prefetch could say there is no issue for a user, then another process creates and emails an issue for that user before this task gets to them. When this task eventually processes the user it trusts the stale miss and creates another issue. `send_issue()` does re-read the flags, but it does so for the new issue ID. That new issue has not been emailed, so as far as I can see we can still end up sending the certificate twice.

I know duplicate issue rows are currently possible because there is no unique constraint, but I don't think we should deliberately introduce another path that can create them, especially if that can also result in another email being sent.

Could you have another look at this case and add a test that reproduces something like: prefetch says no issue exists, another process creates/emails the issue, then the original process continues using its stale prefetch? I think the important thing is that we shouldn't create/send another issue in that situation.

I also wonder if treating a missing key in `$existingissues` as authoritative is a little dangerous. The method itself does not know whether that user was actually part of the original prefetch, so an incomplete array from another place would also mean "this user definitely has no issue". Maybe the structure could explicitly represent users that were checked and had no issue, rather than using a missing key to mean that.

Also, can you please squash this into one commit and address the remaining issues found by GHA? Thanks! :)

